PR #153, opened 2013-07-03
Streaming replication is pathologically slow

rvagg:
Streaming one level instance into another is much slower than expected. @dominictarr was looking into this with me at Nodeconf using this gist https://gist.github.com/brycebaril/5893248

The gist has two files, one that creates a small 1e6 record leveldb instance (~24MB on disk) and another which will attempt to use `db1.createReadStream().pipe(db2.createWriteStream())` to clone the instance to another level instance.

What we were seeing is that once the initial batch is sent, each buffered batch will typically be only a single record, significantly slowing down the transfer. In our case it was taking ~75 seconds.

### DIFF
--- a/lib/write-stream.js
+++ b/lib/write-stream.js
@@ -13,7 +13,11 @@ var Stream       = require('stream').Stream
 
   , getOptions   = require('./util').getOptions
 
-  , defaultOptions = { type: 'put' }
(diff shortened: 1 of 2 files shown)


ralphtheninja:
@mcollina Yes we should :)

PR #187, opened 2013-09-10
Add typewise encoding option to levelup.

mikeal:
This ads encoding support for CouchDB/indexeDB style type sorting via bytewise.

https://github.com/deanlandolt/typewise

The reason this should go in to levelup rather than being in userland above levelup using either binary or ascii encoding is because it turns out to be pretty expensive and rather painful to encode/decode in every sublevel or equivalent byteslice portion of a levelup.

@dominictarr has also been talking about moving sublevel to use arrays and typewise sorting for which I think this patch is probably a requirement.

It should also be noted that the feature is called "typewise" as it should be implementation agnostic as to the actual encoder so that in the future we might be able to optimize the encoder/decoder for different implementations (CouchDB and IndexedDatabase would essentially be noops).

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -7,6 +7,7 @@
 var extend = require('xtend')
   , errors = require('./errors')
   , bops   = require('bops')
+  , bytewise = require('bytewise')
 
   , encodingNames = [
         'hex'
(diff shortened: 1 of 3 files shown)


ralphtheninja:
@mikeal I tried going through this but it seems the repository from where the PR originates from is no more. This PR is 18 months old and the last comment was made a year ago. Please submit a new PR if you feel this is still needed.

PR #207, opened 2013-10-14
Remove write-stream from levelup!

### LINKED ISSUE #199: Bite the bullet and remove WriteStream completely?
WriteStream is only there for the sake of symmetry but that's not a great reason in itself if there are better reasons for it not to be there. I think we're starting to realise that there are different use-cases that require different implementations and we're already seeing some alternative implementations so why not set them free to coexist and compete in user-land?

Discuss

If we conclude that it should be removed then we can add documentation to the ReadStream section saying that if you want a WriteStream then go to a particular place on the wiki to see a list of implementations and what they are good for. We could even dedicate a whole wiki page to this if need be.

if we concluded that it should stay, then I'll try and get these competing pull requests for new WriteStreams reconciled and sort something out because having a streams2 ReadStream and a streams1 WriteStream is a mess and this needs to be resolved.

jcrugzz:
Thought I would go ahead and remove all the references to rip off the
bandaid. @rvagg let me know if i missed something but all the tests
pass. Closes #199


ralphtheninja:
@jcrugzz I'm closing this PR and will continue working on your rebased branch (that lives in `levelup` now)

PR #289, opened 2014-10-19
1.0.0-wip

kesla:
Let's get a 1.0.0 done to compliment leveldown@1.0.0

- [x] Update leveldown to 1.0.0
- [x] Allow saving of empty values
- [x] Remove write-stream (#207)

Have I missed something else we should try and get into this release?

PR #296, opened 2014-12-25
remove WriteStream, replace in tests with level-ws

rvagg:
PR is against 1.0.0-wip / #289 / @kesla 

I've removed write-stream.js but have left the tests intact by using level-ws but it'll depend on 1.0.0 if & when Level/level-ws#2 is merged.

The only part missing is the fstream stuff for functional tests. I'll work on that next.


ralphtheninja:
lgtm

I have tested this code with this PR https://github.com/Level/level-ws/pull/2. I suggest we merge it and release a new version of `level-ws` and add update that dependency into this PR.

Thoughts?

ralphtheninja:
Oh right. The functional tests aren't passing. Missed that.

PR #299, opened 2015-01-12
levelup from already-created leveldown

substack:
The documentation would seem to suggest that you can pass in an `opts.db` leveldown object:

* `'db'` *(object, default: LevelDOWN)*: LevelUP is backed by [LevelDOWN](https://github.com/rvagg/node-leveldown/) to provide an interface to LevelDB.

It also mentions that db can be a constructor, but it doesn't mention that this is the only way to create a levelup from a leveldown. Currently if you already have a leveldown instance you've got to do:

``` js
var levelup = require('levelup');
var leveldown = require('leveldown');

var down = leveldown('./whatever.db');
var up = levelup({ db: function () { return down });
```

which is pretty strange since levelup instances are 1:1 with leveldown instances. I really don't see the benefit in passing a constructor function at all. It seems like it should be levelup's job to just upgrade existing leveldown instances, not to instantiate leveldown instances from a constructor with a location.

This patch makes the following work like the documentation suggests it should work:

``` js
var levelup = require('levelup');
var leveldown = require('leveldown');

var down = leveldown('./whatever.db');
var up = levelup({ db: down });
```

The documentation also says:

``` md
### levelup(db[, callback ])
```

which would appear to suggest that you can do `levelup(down)`, but that doesn't actually work.
This patch also adds support for this already-documented form:

``` js
var levelup = require('levelup');
var leveldown = require('leveldown');

var down = leveldown('./whatever.db');
var up = levelup(down);
```

or you could just do:

``` js
var levelup = require('levelup');
var leveldown = require('leveldown');

var db = levelup(leveldown('./whatever.db'));
```

These forms to me seem much closer to the intent of levelup: to wrap features such as encodings and createReadStream on top of a leveldown.

Further out, I'm not really sure what the location property is doing in there at all or what it's for. It seems like handling a location path should be leveldown's job, not levelup. Plus there are more ways to specify a "location" than a simple string value, which leveldowns will be better at providing given their unique specializations.

### DIFF
--- a/lib/is-leveldown.js
+++ b/lib/is-leveldown.js
@@ -1,0 +1,7 @@
+module.exports = function (db) {
+  if (!db || typeof db !== 'object') return false
+  return typeof db.iterator === 'function'
+    && typeof db.get === 'function'
+    && typeof db.put === 'function'
+    && typeof db.batch === 'function'
+}
--- a/lib/levelup.js
+++ b/lib/levelup.js
@@ -22,6 +22,7 @@ var EventEmitter   = require('events').EventEmitter
   , util           = require('./util')
   , Batch          = require('./batch')
   , codec          = require('./codec')
+  , isLevelDOWN    = require('./is-leveldown')
 
   , getOptions     = util.getOptions
   , defaultOptions = util.defaultOptions
@@ -45,17 +46,27 @@ function LevelUP (location, options, callback) {
   if (!(this instanceof LevelUP))
     return new LevelUP(location, options, callback)
 
-  var error
+  var error, down
 
   EventEmitter.call(this)
   this.setMaxListeners(Infinity)
 
-  if (typeof location == 'function') {
+  if (isLevelDOWN(location)) {
+    options = typeof options == 'object' ? options : {}
+    down = location
+    options.db = function () { return down }
+    location = null
+  } else if (typeof location == 'function') {
     options = typeof options == 'object' ? options : {}
     options.db = location
     location = null
   } else if (typeof location == 'object' && typeof location.db == 'function') {
     options = location
+    location = null
+  } else if (typeof location === 'object' && isLevelDOWN(location.db)) {
+    options = location
+    down = options.db
+    options.db = function () { return down } 
     location = null
   }
 
--- a/test/levelup-from-leveldown.js
+++ b/test/levelup-from-leveldown.js
@@ -1,0 +1,55 @@
+/* Copyright (c) 2012-2014 LevelUP contributors
+ * See list at <https://github.com/rvagg/node-levelup#contributing>
+ * MIT License <https://github.com/rvagg/node-levelup/blob/master/LICENSE.md>
+ */
+
+var levelup = require('../lib/levelup.js')
+  , assert  = require('referee').assert
+  , refute  = require('referee').refute
+  , buster  = require('bustermove')
+  , MemDOWN = require('memdown')
+
+require('./common')
+
+buster.testCase('levelup from leveldown', {
+    'levelup from leveldown in location slot': function (done) {
+      var md       = new MemDOWN('foo')
+        , db       = levelup(md)
+        , entries  = []
+        , expected = [
+              { key: 'a', value: 'A' }
+            , { key: 'b', value: 'B' }
+            , { key: 'c', value: 'C' }
+            , { key: 'd', value: 'D' }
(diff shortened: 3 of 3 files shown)


rvagg:
+1 on this from me

`location` is an awkward quirk of history that we still haven't managed to get away from, perhaps it's time we completely ripped leveldown-by-default out of levelup and left that job to the level package, that way we don't need to do any initialisation or pass anything to a new leveldown object, we can just _require_ a leveldown be passed in here.

For my own use, I almost always use *level* when I just want a vanilla leveldown+levelup setup.

dominictarr:
how does this handle open/closed state. I think the reason that levelup expects a constructor function is so that it can manage open state, so that it's possible to get a sync open ui. Does this still work if you pass in an already opened leveldown?

juliangruber:
:+1:

mcollina:
:+1: for me.

substack:
@dominictarr: that does seem to be a problem:

``` js
var levelup = require('./');
var leveldown = require('leveldown');

var rows = [
    { type: 'put', key: 'a', value: 555 },
    { type: 'put', key: 'b', value: [ 3, 4, 5 ] },
    { type: 'put', key: 'c', value: { x: 7, y: 8 } }
];

var down = leveldown('/tmp/cool.db');
down.open(function () {
    console.log('open!');
    
    var up = levelup(down, { valueEncoding: 'json' });
    up.batch(rows, function (err) {
        if (err) return console.error(err);
        up.createReadStream().on('data', console.log);
    });
});
```

gives this output:

```
 $ node ex.js
open!

events.js:72
        throw er; // Unhandled 'error' event
              ^
OpenError: IO error: lock /tmp/cool.db/LOCK: already held by process
    at /home/substack/projects/node-levelup/lib/levelup.js:129:34
```

substack:
This patch seems to fix the lock error above, but I'm not sure if there are other implications:

``` patch
From 39784034c68f561b88fc0ba01a4c1c1a4d3daa9e Mon Sep 17 00:00:00 2001
From: James Halliday <mail@substack.net>
Date: Tue, 13 Jan 2015 00:39:53 -0800
Subject: [PATCH] don't try to open a leveldown

---
 lib/levelup.js |    7 ++++++-
 1 file changed, 6 insertions(+), 1 deletion(-)

diff --git a/lib/levelup.js b/lib/levelup.js
index e48e31d..10783b9 100644
--- a/lib/levelup.js
+++ b/lib/levelup.js
@@ -94,7 +94,12 @@ function LevelUP (location, options, callback) {
   // set this.location as enumerable but not configurable or writable
   prr(this, 'location', location, 'e')
 
-  this.open(callback)
+  if (down) {
+    this._status = 'open'
+    this.db = down
+  } else {
+    this.open(callback)
+  }
 }
 
 inherits(LevelUP, EventEmitter)
-- 
1.7.10.4

```

The documentation in leveldown is quite clear that leveldown does not track the open states, so it might be good to add a reminder to the levelup docs that any leveldown that is passed in should already be open.

juliangruber:
the cleaner solution seems to be to add a `.open` state to leveldown, so you can pass open and closed leveldown instances. however given our ecosystem, the approach above also sounds good to me

dominictarr:
I agree with julian here, to do this properly we need a state property on leveldown.

ralphtheninja:
I'd like to see this functionality in `1.0.0`. I suggest we merge `1.0.0-wip` branch into master first and then rebase this ontop of latest master. @substack sound ok?

kesla:
+1 Let's get this into levelup 1.0!

jcrugzz:
im +1 on this if it wont block a better implementation (keeping the open/closed state on leveldown). @kesla you know the implications here?

ralphtheninja:
Rebase rebase rebase! :)

kesla:
@jcrugzz sorry no

ralphtheninja:
> Further out, I'm not really sure what the location property is doing in there at all or what it's for. It seems like handling a location path should be leveldown's job, not levelup. Plus there are more ways to specify a "location" than a simple string value, which leveldowns will be better at providing given their unique specializations.

I agree. Lets get rid of `location` in a separate PR.

ralphtheninja:
Moved this to new PR. Closing. Thank you @substack !

ralphtheninja:
And +1 for having `.open` state in `leveldown`

PR #302, opened 2015-01-19
updates for iojs (WIP)

ralphtheninja:
This makes `node-levelup` work with latest `node-leveldown`. However, we also need to release a new version of `deferred-leveldown` and update dependencies before merging this into master.

See https://github.com/Level/deferred-leveldown/pull/3


ralphtheninja:
`batch.put()` tests working now. Having some problems with the functional tests.

ralphtheninja:
Just realized I'm working on the wrong branch. Rebasing this onto `1.0.0-wip` instead.

PR #305, opened 2015-01-20
update logo and copyright

ralphtheninja:
The logo in this README worked but I changed it to the one we are using on `leveldb.org`. Also I think it's pretty cool to link to a logo from a site we are hosting. Also makes it easier to replace it later with another svg if we like.


juliangruber:
+1

PR #307, opened 2015-02-14
Added errors to the available namespace when requiring 'levelup'.

braydonf:
This adds the errors to the available namespace so that it can be used from `levelup`.

```javascript
var levelup = require('levelup');
var db = levelup('./somedata');

db.get('somekey', function(err, value) {
  if (err instanceof levelup.errors.NotFoundError) {
    return db.put('somekey', 'somedata', function (err) {
      if (err) {
        throw err;
      }
    });
  } else if (err) {
    throw err;
  }
});
```

### DIFF
--- a/lib/levelup.js
+++ b/lib/levelup.js
@@ -449,6 +449,7 @@ function utilStatic (name) {
 }
 
 module.exports         = LevelUP
+module.exports.errors  = require('./errors');
 // DEPRECATED: prefer accessing LevelDOWN for this: require('leveldown').destroy()
 module.exports.destroy = utilStatic('destroy')
 // DEPRECATED: prefer accessing LevelDOWN for this: require('leveldown').repair()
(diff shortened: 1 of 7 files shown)


ralphtheninja:
:+1:

kesla:
For `NotFoundError` you can do

```js
if (err.notFound) {
  ...
}
```

juliangruber:
lgtm, needs docs as well, and maybe a blog post?

rvagg:
sure, may as well put this in the v1.0 bucket

ralphtheninja:
@braydonf Mind rebasing this onto latest master?

braydonf:
Rebased

jcrugzz:
+1 on this. Seems like a candidate for its own module but this isn't a bad stepping stone

ralphtheninja:
@jcrugzz Aye. We'll get there :)

ralphtheninja:
@braydonf Thanks!

PR #312, opened 2015-03-18
extract error codes into level-errors module

ralphtheninja:


jcrugzz:
:+1: from me. Allows other `level` modules to better reuse `level` errors without directly depending on `levelup`

mcollina:
:+1:, but what exactly is the use case for this?

ralphtheninja:
@mcollina What @jcrugzz said.

@juliangruber started with `codec` yesterday and I thought I'd follow a long in the spirit of this issue https://github.com/rvagg/node-levelup/issues/270

PR #313, opened 2015-03-18
Use level-codec

juliangruber:
This isn't yet ready to be merged, due to lacking sability in `level-codec`.

I initially factored out the encoding logic so I could use it in multilevel, but this also helps slim down levelup's code / responsibilities.

I'm thinking about moving more logic into level-codec later, in a separate pull request, like wrapping the callback for get / put and doing the encoding behind the scenes, instead of explicitly doing it in levelup.

What do you think?

__As a bonus, I removed `{encoding}` which should be merged before `1.0.0` is published.__


juliangruber:
oops, forgot to publish latest level-codec, see now: https://github.com/Level/codec/blob/master/index.js. The implementatino is a draft still with the purpose of having something that works so tests pass.

About passing in the codec: That neither was tested nor documented, so it looked like noone was using it.

juliangruber:
tests pass now

mafintosh:
does this mean i can't do this anymore?

``` js
db.get('foo', {valueEncoding: myOwnEncoder}, cb)
```

i usually do that when using protocol-buffers, https://github.com/mafintosh/protocol-buffers#leveldb-encoding-compatibility

juliangruber:
you totally can! in fact there's tests already in levelup to prove it. this doesn't change the external api at all, it's just an internal refactoring.

mafintosh:
@juliangruber :+1:

juliangruber:
well, the only thing to change the internal api is the deprecation of `.encoding`, but there's a separate pull request to discuss this

ralphtheninja:
LGTM

ralphtheninja:
@juliangruber Mind squashing commits before merge?

juliangruber:
I will, once level-codec is ready (read: works in multilevel as well, has stable api and more tests), and maybe waiting for more :+1:s

jcrugzz:
LGTM once you think `level-codec` is GTG @juliangruber. :+1: all the modularization.

juliangruber:
:dancer:

PR #314, opened 2015-03-19
remove encoding option

juliangruber:
We undocumented it a while I go, now it's time to finally kick it out for `1.0.0`.

### DIFF
--- a/lib/levelup.js
+++ b/lib/levelup.js
@@ -312,7 +312,7 @@ LevelUP.prototype.batch = function (arr_, options, callback) {
 
     // inherit encoding
     var kEnc = e.keyEncoding || keyEnc
-      , vEnc = e.valueEncoding || e.encoding || valueEnc
+      , vEnc = e.valueEncoding || valueEnc
       , o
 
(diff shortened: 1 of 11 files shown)


lib/levelup.js
```diff
@@ -377,8 +377,8 @@ LevelUP.prototype.createReadStream = function (options) {
   var self = this
   options = extend( {keys: true, values: true}, this.options, options)
 
-  options.keyEncoding   = options.keyEncoding   || options.encoding
```
juliangruber: this was bad!

juliangruber: should we add a test for this? `.encoding` was always only meant as an alias for `.valueEncoding`

ralphtheninja: What do you want to test?

juliangruber: oh yeah right no need to test this any more

jcrugzz:
I'm :+1: on this as well. Too ambiguous and since it wasn't documented, it fits well for `1.0.0`

PR #315, opened 2015-03-20
remove reference to write-stream and iterators

ralphtheninja:
I'm also wondering if we should try to reformulate a bit. Since I'm not a native english speaker I might be off. In the beginning we mention `bi-directional iterators` as one of the features and we say this in plural, i.e. there are at least two types of iterators.

But since `write-stream` has been removed, does this mean we have one type of iterator, or are there still several types of iterators? Or is it a single bi-directional iterator that is exposed by `createReadStream()`? If there are different types of iterators I think we should be more specific in what they are to get rid of ambiguity.


ralphtheninja:
@juliangruber Any thoughts?

jcrugzz:
@ralphtheninja this seems fine. The `bi-directional` iterators in this case means that they can go either forward or reverse.

PR #318, opened 2015-03-22
Changed options for get to same as put

RichardLitt:
The `get` options paragraph didn't mention that `value` objects also adhrere to `valueEncoding`. I just removed the duplicate paragraph with less information and stated that `get` is the same as `put` - seems more DRY.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ db.get('foo', function (err, value) {
 
 #### `options`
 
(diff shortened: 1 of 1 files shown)


ralphtheninja:
:+1: Mind changing it for `del` as well?

ralphtheninja:
:+1:

ralphtheninja:
Perhaps don't mention the value object, since it's not related to `get`.

> Encoding of the `key` object is the same as in <a href="#put"><code>put</code></a>.

ralphtheninja:
Thanks!

PR #319, opened 2015-03-23
fix benchmarks by installing leveldown@^0.10.4

juliangruber:
so the old levelup uses a version of leveldown it accepts.

i just noticed that this isn't important to merge since 1.0.0 will be out soon and that will become the version of `levelup-release`.


ralphtheninja:
:+1: But should still use leveled? It seems to be deprecated in favor of `level` and also it doesn't work with iojs.

juliangruber:
yeah, let's remove leveled!

ralphtheninja:
Works like a charm on iojs!

PR #320, opened 2015-03-29
refactor iterators using new deferred-leveldown

juliangruber:
this depends on https://github.com/Level/deferred-leveldown/pull/6

This removes deferred iterator logic from levelup as that's handled in the pull request above. Also the readable stream interface is one step closer to being ready to be split out into its own module.

### DIFF
--- a/lib/levelup.js
+++ b/lib/levelup.js
@@ -140,7 +140,7 @@ LevelUP.prototype.close = function (callback) {
         callback.apply(null, arguments)
     })
     this.emit('closing')
-    this.db = null
+    this.db = new DeferredLevelDOWN(this.location)
   } else if (this._status == 'closed' && callback) {
     return process.nextTick(callback)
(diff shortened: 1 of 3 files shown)


lib/levelup.js
```diff
@@ -140,7 +140,7 @@ LevelUP.prototype.close = function (callback) {
         callback.apply(null, arguments)
     })
     this.emit('closing')
-    this.db = null
+    this.db = new DeferredLevelDOWN(this.location)
```
juliangruber: go back to the initial state

juliangruber: this was required for tests to pass, removing `this.db = null` would result in a segfault

ralphtheninja:
:+1:

juliangruber:
once this is merged, use [level-iterator-stream](https://github.com/Level/iterator-stream) to refactor

PR #324, opened 2015-04-17
remove .approximateSize()

juliangruber:
Since it has been undocumented for a while and we're about to make a major release.

We can also use [util.deprecate](https://iojs.org/api/util.html#util_util_deprecate_function_string) first, but I'd prefer plain removing it.


ralphtheninja:
@juliangruber on a roll today! :)

juliangruber:
i decided for the more resonable way of using `util.deprecate` first. we can then remove `.approximateSize` in half a year or so.

PR #326, opened 2015-04-20
fix stream-bench.js

juliangruber:
This needed to be touched since we removed writestreams. i also added benchmarks for different `{ keys, values }` combinations and made sure they run often enough so performance is a stable characteristic.

Current output:

```bash
 ∴  benchmarks (master) : node stream-bench.js
Filled source! Took 2702ms, reading data now...
Done! Took 1682ms with {}
Done! Took 1336ms with {"keys":true}
Done! Took 1083ms with {"values":true}
Done! Took 1222ms with {"keys":true,"values":true}
Done! Took 1112ms with {}
Done! Took 1134ms with {"keys":true}
Done! Took 1065ms with {"values":true}
Done! Took 1182ms with {"keys":true,"values":true}
Done! Took 1272ms with {}
Done! Took 1086ms with {"keys":true}
Done! Took 1106ms with {"values":true}
Done! Took 1416ms with {"keys":true,"values":true}
```

The need for stream benchmarks was raised again in https://github.com/Level/codec/pull/5#issuecomment-94213135

### DIFF
--- a/test/benchmarks/stream-bench.js
+++ b/test/benchmarks/stream-bench.js
@@ -1,7 +1,6 @@
 const levelup = require(process.argv[2] || '../../')
     , crypto  = require('crypto')
     , srcdb   = levelup('/tmp/source.db')
-    , dstdb   = levelup('/tmp/destination.db')
 
     , batch   = 10000
     , total   = 200000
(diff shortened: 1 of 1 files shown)


ralphtheninja:
:+1:

PR #332, opened 2015-04-28
update dependencies

ralphtheninja:
Most important thing here imo is `deferred-leveldown@1.0.0` (`deferred-leveldown` should use the same `abstract-leveldown` as `leveldown` does). Only outdated module now is `tap`. I can't update to latest because npm complains when installing `bustermove`, but I don't think that matters that much.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,13 +34,13 @@
   ],
   "main": "lib/levelup.js",
   "dependencies": {
-    "deferred-leveldown": "^0.3.0",
-    "level-codec": "^5.0.0",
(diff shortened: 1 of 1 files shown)


mcollina:
How about moving out of tap in favor of tape?

ralphtheninja:
@mcollina I'm all for that!

juliangruber:
+1 for tape over tap

ralphtheninja:
Just need https://github.com/rvagg/bustermove/pull/2 to land and then we can switch to `tape`

PR #334, opened 2015-05-06
target multiple iojs versions, remove notifications

ralphtheninja:
Also cut testing time in half.

Before:

levelup-travis-before

After:

levelup-travis-after

/cc @juliangruber


juliangruber:
n n n n niiiiice!

PR #338, opened 2015-05-08
Guard to call "close" after "get", "put", "del" and "batch"

braydonf:


braydonf:
Updated from feedback.

braydonf:
`guardCallback` has been updated

jcrugzz:
Not a bad thing to have to ensure safety, I'd be curious if it causes any perf issues but it is only 1 extra function call

braydonf:
Running the benchmarks with release as levelup@1.0.0-5 and leveldown@1.0.6:
```
LevelUP           put(int, string) x 10 x 659 ops/sec ±1.40% (44 runs sampled)
LevelUP (release) put(int, string) x 10 x 647 ops/sec ±1.02% (77 runs sampled)

LevelUP           put(int, string) x 1000 x 136 ops/sec ±3.70% (68 runs sampled)
LevelUP (release) put(int, string) x 1000 x 161 ops/sec ±3.15% (81 runs sampled)

LevelUP           put(int, string) x 100,000 x 1.73 ops/sec ±5.15% (13 runs sampled)
LevelUP (release) put(int, string) x 100,000 x 1.70 ops/sec ±9.16% (12 runs sampled)

LevelUP           get(int):string x 10 x 664 ops/sec ±1.97% (37 runs sampled)
LevelUP (release) get(int):string x 10 x 664 ops/sec ±2.04% (41 runs sampled)

LevelUP           get(int):string x 1000 x 124 ops/sec ±0.76% (69 runs sampled)
LevelUP (release) get(int):string x 1000 x 131 ops/sec ±0.90% (79 runs sampled)

LevelUP           batch(int, string) x 1000 x 201 ops/sec ±3.63% (60 runs sampled)
LevelUP (release) batch(int, string) x 1000 x 209 ops/sec ±3.21% (62 runs sampled)

LevelUP           batch(int, string) x 100,000 x 2.60 ops/sec ±2.28% (17 runs sampled)
LevelUP (release) batch(int, string) x 100,000 x 2.69 ops/sec ±0.84% (18 runs sampled)
```

braydonf:
Updated. BTW: jshint isn't giving any warnings about the style changes, should it?

jcrugzz:
@ralphtheninja we should think about making a `level-styleguide` with something like https://www.npmjs.com/package/fashion-show

ralphtheninja:
@jcrugzz Awesome idea. Feel like hacking together one? :)
@braydonf It should :)

lib/levelup.js
```diff
@@ -132,7 +129,9 @@ LevelUP.prototype.open = function (callback) {
 LevelUP.prototype.close = function (callback) {
   var self = this
 
-  if (this.isOpen()) {
+  if (this._activeOperations > 0) {
+    this.on('inactive', this.close.bind(this, callback))
```
juliangruber: should be a `.once`, because you can close and open a db more than once

juliangruber:
for a complete solution this needs to guard against creation of new streams as well

juliangruber:
since leveldown takes care of closing iterators when you `.close()`, maybe it should take care of the async functions as well. see https://github.com/Level/leveldown/blob/master/test/cleanup-hanging-iterators-test.js#L32

hmmm my gut tells me this should all either go into levelup or abstract-leveldown, so none of the backends needs to take care of this stuff. what do you think?

lib/levelup.js
ralphtheninja: :+1:

braydonf:
Initially I thought the best place would be in `Database::Close` in leveldown, similar to the [iterators check](https://github.com/Level/leveldown/blob/master/src/database.cc#L263), however after running gdb with levelup code similar to:
```js
db.put('key', 'value', function() {});
db.close(function() {});
```
`Database::Close` was called and completed *first*, even though put was called. So having the guard as close to the caller and the original order makes sense.

Though I still think the backends should handle it to some degree, i.e. in the case of leveldown, if `Database::Put` is called after `Database::Close` it should [give an error](https://github.com/Level/leveldown/issues/157#issuecomment-100020704) rather than crash with a segfault.

ralphtheninja:
Closing due to outdated and too much stuff has happened lately to be able to merge this.

PR #342, opened 2015-05-18
Levelup from leveldown

ralphtheninja:
Re-creating https://github.com/Level/levelup/pull/299 as a new PR with cherry picked commits. Also using `isLevelDOWN()` from `abstract-leveldown`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,24 @@ var memdown = require('memdown')
 var db = levelup(memdown)
 ```
 
+Both `levelup(options, callback)` and `levelup(db, callback)` can be called with an already created LevelDOWN API compatible object. The database object must not have been opened (see below).
+
+For example:
+
+```js
+var levelup = require('levelup')
+var leveldown = require('leveldown')
+var db = levelup({ db: leveldown('./mydb') })
+```
+
+or
+
+```js
+var levelup = require('levelup')
+var leveldown = require('leveldown')
+var db = levelup(leveldown('./mydb'))
+```
+
 #### `options`
 
 `levelup()` takes an optional options object as its second argument; the following properties are accepted:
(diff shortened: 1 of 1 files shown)


lib/levelup.js
```diff
@@ -47,20 +47,30 @@ function LevelUP (location, options, callback) {
...
+    , down
 
   EventEmitter.call(this)
   this.setMaxListeners(Infinity)
 
-  if (typeof location == 'function') {
+  if (isLevelDOWN(location)) {
+    options = typeof options == 'object' ? options : {}
+    down = location
+    options.db = function () { return down }
```
juliangruber: why is that still a function?  i think now is a good time to enable `options.db = down`

juliangruber: ah I see in the tests that that already works, so we can simplify here?

juliangruber:
given https://github.com/Level/levelup/pull/299#issuecomment-69700570, we should document that the leveldown must not have been opened

lib/levelup.js
ralphtheninja: Yes. This if-statement looks terrible :) Imo the best would be to remove the `location` parameter completely.

juliangruber: Shall me make it only this:

```js
var db = levelup(leveldown(location))
```

ralphtheninja: Skip the `callback` completely?

ralphtheninja: I'm all for simplifying, but I'd prefer to do all of that in a separate PR. I guess this change will involve a new major?

ralphtheninja:
Agree. I'll update README.

ralphtheninja:
Locking PR until leveldown can keep track of its own open state.

PR #346, opened 2015-05-26
Remove implicit leveldown

juliangruber:
Implicit leveldown is based on two assumptions:

- leveldown is the primary backend for levelup
- people use levelup directly (instead of convenience packages like level)

I consider both not true any more, so it's time to clean up for v2.

This pull request is based on the `remove/deprecated` branch, since without implicit leveldown it's impossible to expose `levelup.repair` and `levelup.destroy`.


mcollina:
+1
Il giorno mar 26 mag 2015 alle 19:41 Julian Gruber <notifications@github.com>
ha scritto:

substack:
This looks great! And we can remove all the path handling logic too that was clogging up the `levelup(leveldown(...))` use-case (which should be the preferred use-case now).

jcrugzz:
+1. Seems like this will go hand in hand with open state being handled in leveldown?

juliangruber:
> This looks great! And we can remove all the path handling logic too that was clogging up the levelup(leveldown(...)) use-case (which should be the preferred use-case now).

exactly, this is part of preparation for that. i have a `julian-v2` branch https://github.com/Level/levelup/tree/julian-v2 that has that already implemented, and I'm going to try to put this into master for v2 one step at a time.

juliangruber:
> +1. Seems like this will go hand in hand with open state being handled in leveldown?

yup! abstract-leveldown already got .status [here](https://github.com/Level/abstract-leveldown/commit/f5ac25da9625af540b08cdf6b4f1383726b8daeb), now we just need to figure out the sane way for levelup to use it

ralphtheninja:
Good work!

PR #350, opened 2015-05-29
Remove location argument

juliangruber:
this changes the api from

```js
var db = levelup(location, { db: leveldown })
```

to

```js
var db = levelup(leveldown(location))
```

this is for v2

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ var levelup   = require('levelup')
 
 // 1) Create our database, supply location and options.
 //    This will create or open the underlying LevelDB store.
-var db = levelup('./mydb', { db: leveldown })
+var db = levelup(leveldown('./mydb'))
 
 // 2) put a key & value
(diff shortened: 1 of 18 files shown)


juliangruber:
a lot of code could be simplified, and documentation shortened. now there's only one way to create a db

ralphtheninja:
:+1:

juliangruber:
instead of relying on the package.json to check whether the leveldown is compatible, i added a check whether `.status` exists, which was added recently

juliangruber:
can i get any other feedback?

ralphtheninja:
cc @Level/owners

rvagg:
:thumbsup: it's felt awkward since we split levelup into a generic frontend

ralphtheninja:
w00t

mafintosh:
this is great. :+1:

PR #353, opened 2015-06-09
improve error message when trying to require leveldown

ralphtheninja:
I'm trying to iron out some problems we get when trying to require `leveldown` (see https://github.com/Level/leveldown/issues/190).

Currently we give the advice to try and `npm install leveldown` if requiring it doesn't work. This error message isn't very helpful when you before receiving the error actually did install it.

This PR uses the error message we get in `try/catch` so we can get more information (e.g. the current ABI version and the expected ABI version).

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -4,8 +4,9 @@
  * <https://github.com/level/levelup/blob/master/LICENSE.md>
  */
 
-var extend        = require('xtend')
-  , LevelUPError  = require('level-errors').LevelUPError
+var extend         = require('xtend')
+  , LevelUPError   = require('level-errors').LevelUPError
+  , format         = require('util').format
   , defaultOptions = {
         createIfMissing : true
       , errorIfExists   : false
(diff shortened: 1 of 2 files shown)


ralphtheninja:
Anyone against this? cc @juliangruber

lib/util.js
```diff
@@ -51,10 +51,15 @@ function getLevelDOWN () {
...
     return leveldown = require('leveldown')
   } catch (e) {
-    throw new LevelUPError(missingLevelDOWNError)
+    throwRequireError(e)
   }
 }
 
+function throwRequireError (e) {
+  var template = 'Failed to require LevelDOWN (%s). Try `npm install leveldown` if it\'s missing'
+  throw new LevelUPError(format(template, e.message))
```
juliangruber: IMO the `throw` should still be outside the function, ie it should only return the error

juliangruber:
otherwise lgtm!

lib/util.js
ralphtheninja: Good call. I'll update.

ralphtheninja:
Updated.

ralphtheninja:
```
$ npm publish
+ levelup@1.2.1
```

Issue #365, opened 2015-09-24
move encodings into leveldown


dominictarr:
from discussion on this twitter thread: https://twitter.com/jedschmidt/status/646679054326865920

what would be the pros and cons of just removing encodings from levelup entirely, and making that just something that happens in leveldown? This would allow some backends to use native support for this,
and does anyone _really_ use different encodings for different keys? That would be a huge headache.

This would be a breaking change for all my stuff, in that I'd have to update that code, but it wouldn't break the design because I have never used more than one encoding anyway.

This would be a huge breaking change, but would also simplify a lot, killing loads of code.

I'm not proposing this as a serious change right now, but I do think this would be a possibility that is worth discussing.

juliangruber:
for reference, encodings could be handled inside a leveldown, or inside a "fake" leveldown, like so:

```js
var db = leveldown('./db')
db = encode('buffer', db)
db = levelup(db)

// or

var db = levelup(encode('buffer', leveldown('./db')))
```

juliangruber:
also, we have a `v2` branch already with a couple breaking changes, so this change could fit in there

juliangruber:
also +1 for dropping `.{key,value}Encoding` from any operation exposed in levelup

mcollina:
I'm :+1: for having a single encoding for each database. We should support different encodings for keys and values.

For this to happen, I think we need:

1. an easy way for adding custom encodings to the current *down implementations, possibly as part as levelup api (in the constructor options)
2. a strategy to maintain the current 1.x.x series for 6-12 months (leveldown nan bumps)

jcrugzz:
I'm ok with this. The most I've used encoding wise is `utf8` and `json` values but they were in separate `sublevels` using @mafintosh's `subleveldown`. If my understanding is correct, since `subleveldown` wraps each `sublevel` with a `leveldown`, this will work as expected. The API just might have to change internally slightly.

juliangruber:
pull request: https://github.com/Level/levelup/pull/367

ralphtheninja:
I'm in huge favour of going in this direction and see several advantages:

* simplified code base -> a lot less can go wrong which means less bugs and less work to maintain it, increased readability and all that good stuff
* levelup assumes less -> leveldowns have more freedom to do what they want as long as they support the tests of abstract-leveldown
* levelup can be *done*
* potentially also improvements to performance since encodings could also be pushed to the c++ layer (if there is one obviously)

dominictarr:
[...]

PR #367, opened 2015-09-25
remove encodings

juliangruber:
As suggested in https://github.com/Level/levelup/issues/365. This pull request is targeted at the `v2` branch.
- removes all notions of any encoding
- leveldown can return any thing, only in the test suite assumes leveldown will always return a buffer

TODO
- [x] write a POC leveldown wrapper that supports current levelup@1 encodings (https://github.com/Level/encoding-down)
- [x] maybe use this wrapper in tests in order to undo all the `.toString()` changes and have a cleaner diff
- [ ] use npm version of level-iterator-stream
- [ ] use encoding-down in levelup to evaluate merging this into master branch now

Stats:
- `-54` lines of documentation
- `-60` lines of lib code
- `-1` dependencies (level-codec)
- `~-479` lines of tests

What do you all think?


mcollina:
I'm not sure about

> assumes leveldown will always return a buffer

If encodings are handled at leveldown side, it can return anything.

juliangruber:
@mcollina ah, i was only talking about the tests. i'll update the pr description

ralphtheninja:
Hmm. Maybe we could extend `abstract-leveldown` with an `encode()` and `decode()` api (with default behavior) + tests for that. Then we could have sanity tests in levelup, like:

```js
db.get('k' + k, function (err, v) {
  assert.equals(v, db.decodeValue(db.encodeValue('v' + k)))
})
```

Does that make sense?

ralphtheninja:
+1 Looks nice btw. Even though there are some coding style discrepancies, but we can clean that up later :)

juliangruber:
> +1 Looks nice btw. Even though there are some coding style discrepancies, but we can clean that up later :)

haha of course :P 

> Hmm. Maybe we could extend abstract-leveldown with an encode() and decode() api (with default behavior) + tests for that. Then we could have sanity tests in levelup, like:
> 
> db.get('k' + k, function (err, v) {
>   assert.equals(v, db.decodeValue(db.encodeValue('v' + k)))
> })
> Does that make sense?

hm I _think_ I see what you're saying...to me it seems having encodings be completely optional and handled in special leveldowns is a more decoupled solution

ralphtheninja:
> hm I think I see what you're saying...to me it seems having encodings be completely optional and handled in special leveldowns is a more decoupled solution

Makes sense. So in that case these tests could basically be removed from levelup since we don't know anything about the returned results. Maybe just move them (or update existing) to tests in abstract-leveldown.

mcollina:
While I'm in favor for a decoupled solution, I think we should take into account the performance of an encoding wrapper.
We already have an abstract-level-down wrapper, so maybe the encoding logic should be in there. But we should have numbers to decide.

Also, we should try to ship level() with the encoding logic anyway, to provide a smooth upgrade path (because level() wraps leveldown).

juliangruber:
> While I'm in favor for a decoupled solution, I think we should take into account the performance of an encoding wrapper.
> We already have an abstract-level-down wrapper, so maybe the encoding logic should be in there. But we should have numbers to decide.

yeah, that's why it's important to write the encoding wrapper next. maybe i'll get that done today

> Also, we should try to ship level() with the encoding logic anyway, to provide a smooth upgrade path (because level() wraps leveldown).

absolutely! each such convenience package should bundle the most appropriate encoder

dcousens:
This sounds fantastic :+1:

juliangruber:
Wrote a leveldown wrapper that uses level-codec to supply levelup@1 encoding logic: https://github.com/Level/encoding-down.

Next: integrate into the test suite as mentioned above.

ralphtheninja:
Replaced by https://github.com/Level/levelup/pull/458

Issue #366, opened 2015-09-25
remove { sync } option


juliangruber:
this is leveldown specific

juliangruber:
together with the removal of encodings it looks like we'll only need options objects to pass down to leveldown

ralphtheninja:
Note to self. Remember to close https://github.com/Level/levelup/pull/368 once this is fixed.

PR #368, opened 2015-09-25
doc: remove notion of leveldown-specific sync option

juliangruber:
This pull request is targeted at the `v2` branch. If wished I can recreate it against `master`, shouldn't matter too much.

As in https://github.com/Level/levelup/issues/366


ralphtheninja:
+1

juliangruber:
@ralphtheninja do you think this should go into master of v2?

ralphtheninja:
Has sync been removed from master?

ralphtheninja:
Oh nevermind, it's leveldown specific. Then I think we could include it on master as well. Sorry for my ignorance :)

lib/levelup.js
```diff
@@ -65,9 +65,17 @@ function LevelUP (location, options, callback) {
     options  = {}
   }
```
ralphtheninja: @juliangruber I'm gonna do it your way here and start off by moving db to `options.db` and once that is done get rid of `location`

lib/levelup.js
```diff
@@ -65,9 +65,17 @@ function LevelUP (location, options, callback) {
...
 
-  if ((!options || typeof options.db != 'function') && typeof location != 'string') {
+  if (!options && typeof options.db != 'function') {
```
ralphtheninja: Going through what you did in this PR and it seems to me that this line is wrong, shouldn't it be

```js
if (!options || typeof options.db !== 'function') {}
```

ralphtheninja: cc @juliangruber

PR #369, opened 2015-10-04
test on all major abi versions

ralphtheninja:


ralphtheninja:
We're technically not needed to test levelup on different abi versions, but it will be symmetric to how leveldown is tested and we will also test that `prebuild --download` works for any given version of levelup, even though leveldown is just a devDependency.

juliangruber:
moar tests running -> moar better

PR #379, opened 2016-01-22
bump semver for nsp fix

ralphtheninja:
`nsp` reports a problem becase `levelup` is using a version of `semver` with a known vulnerability, see https://nodesecurity.io/advisories/31

levelup-semver-nsp

levelup-nsp-fix

I originally started looking at updating `level-sublevel` but there were just too many changes to be made *just* to be able to update `semver`, so I think we should release `0.19.1` instead, and then `level-sublevel` can at least use that version of `levelup` to start with.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - "0.8"
   - "0.10"
   - "0.12"
   - "1.0"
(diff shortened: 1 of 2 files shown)


ralphtheninja:
If I can get a +2 then I'll squash and release this as `0.19.1`. Have a good weekend everyone!

juliangruber:
+9000!

ralphtheninja:
0.19.1

PR #409, opened 2016-11-07
update travis

ralphtheninja:


juliangruber:
👍

.travis.yml
```diff
@@ -16,8 +16,8 @@ before_install:
   - export JOBS=max
 
 node_js:
+  - '7'
   - '6'
   - '5'
```
rvagg: remove 5, its gone

rvagg:
We're going `^` for devDeps now? there's an interesting mix in there now. Doesn't bother me if there's a good reason for it.

.travis.yml
ralphtheninja: Will do.

ralphtheninja:
> We're going ^ for devDeps now? there's an interesting mix in there now. Doesn't bother me if there's a good reason for it.

I updated to ^ for devDeps just to get rid of updating things like tape etc. I can update all to use ^ if that's alright.

.travis.yml
ralphtheninja: And yes, I'll squash later :)

ralphtheninja:
Putting back `.dntrc` with updated node versions. I also want to squash this before merge.

.dntrc
```diff
@@ -0,0 +1,19 @@
+## DNT config file
+## see https://github.com/rvagg/dnt
+
+NODE_VERSIONS="\
+  master   \
+  v7.x-staging  \
+  v6.x-staging  \
+  v4.x-staging  \
```
ralphtheninja: @rvagg I picked these branches because I thought it might be interesting to test on running targets. Feel free to give some input here :)

PR #417, opened 2017-01-26
Let encoding-down handle encodings

juliangruber:
Without a notable performance degradation, this gets rid of a lot of encoding code.

Before we merge this, we should make sure

- [ ] non of the dependencies relies on `db.db` being exactly their leveldown instance
- [ ] `encoding-down` has a good test suite

### DIFF
--- a/lib/batch.js
+++ b/lib/batch.js
@@ -10,19 +10,15 @@ var util          = require('./util')
   , getOptions    = util.getOptions
   , dispatchError = util.dispatchError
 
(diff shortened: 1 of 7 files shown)


ralphtheninja:
@juliangruber Maybe this should be brought in before releasing `2.0.0`? (i.e. new issue on v2 milestone)

juliangruber:
yes i think so

ralphtheninja:
Replaced by https://github.com/Level/levelup/pull/458

PR #422, opened 2017-03-05
Added instructions on encoding/decoding buffers in json

wesleytodd:
Ref: https://github.com/Level/codec/issues/13

I was doing this manually, but @juliangruber pointed me to this module.  Figured it would be nice to document a better way for everyone in the future.  This is my first PR on this project, so let me know if there are any other requirements that I have missed here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -571,6 +571,29 @@ JSON data
 
 You specify `'json'` encoding for both keys and/or values, you can then supply JavaScript objects to LevelUP and receive them from all fetch operations, including ReadStreams. LevelUP will automatically *stringify* your objects and store them as *utf8* and parse the strings back into objects before passing them back to you.
 
+If you are using Buffers in either your key or value, the default behavior of `JSON.stringify` will turn your buffers into a plain object like `{type: "Buffer", value: [/* bytes... */]}`.  Decoding of these does not 
+return the data to a Buffer instance.  If you would like a solution which automatically decodes the Buffers back to their original form, you can use [`json-buffer`](https://github.com/dominictarr/json-buffer) as a
+custom encoding.  This will encode the Buffers to `Base64` strings inside your json, and parsing will return them to their original Buffer form.  Here is an example:
+
+```javasctipt
+var jsonBuffer = require('json-buffer')
+var db = levelup(location, {
+  valueEncoding: {
+      encode : jsonBuffer.stringify
+    , decode : jsonBuffer.parse
+    , buffer : false
+    , type   : 'jsonBuffer'
+  }
+})
+db.put('foo', { buf: Buffer.from('bar') }, (err) => {
+  if (err) return console.error(err)
+  db.get('foo', (err, val) => {
+    if (err) return console.error(err)
+    console.log(val) // <Buffer 66 6f 6f>
+  })
+})
+```
+
 <a name="custom_encodings"></a>
 Custom encodings
 ----------------


README.md
```diff
@@ -571,6 +571,20 @@ JSON data
 
 You specify `'json'` encoding for both keys and/or values, you can then supply JavaScript objects to LevelUP and receive them from all fetch operations, including ReadStreams. LevelUP will automatically *stringify* your objects and store them as *utf8* and parse the strings back into objects before passing them back to you.
 
+If you are encoding buffers and would like a solution which automatically decodes the buffers back to their original form, you can use [`json-buffer`](https://github.com/dominictarr/json-buffer) as a custom encoding like this:
```
vweevers: Can you clarify "encoding buffers", to avoid confusing newcomers? This is about encoding buffers "as Base64 JSON strings", or something like that. It's an important distinction from the existing `binary` encoding.

Perhaps this belongs on the wiki though?

wesleytodd: Better?

vweevers:
Before we start bikeshedding language, I want to explain why I suggested the wiki. We don't currently have a readme section dedicated to binary encoding. That should come first. The documentation on all built-in encodings falls short IMO, so there's no place for custom encodings yet. That'd be an invitation to XY problems.

Additionally, encoding buffers in JSON like this is a rare thing to do (?) and forgoes the performance benefits of reading and writing buffers. Not something to advertise in the readme, perfect for the wiki.

wesleytodd:
Oh sorry, I must have read over that line about the wiki.  I personally am on the fence on that, because with such a fully fleshed out readme already, I didn't even think to look for a wiki.  Secondly, my actual entry point to this was the npm page, so unless there is a strong call out to the wiki many people might miss it.  I personally would rather see additional sections added to the readme over more info in the wiki, but that is just an opinion, take it or leave it :)

I agree that the built in encoding are sparsely documented, but I do not see that as a reason to not start now.  As a new user, I just added this in the section I would have looked for it in, but am happy to move it around however other people think best.

As for how often people want to encode buffers to json, I do not know.  Now that you mention it I might decide to switch to storing protocol buffers with a binary encoding since my app is already utilizing `protobuf`'s.  But that is specific to my case, so I still feel like someone else must have hit this and expected their buffers to be decoded back to buffers.

vweevers:
> so unless there is a strong call out to the wiki many people might miss it

Good point. There's a link near the bottom of the readme, it definitely deserves a more prominent place. Maybe we can put a link in the [Introduction](https://github.com/Level/levelup#introduction), saying "There's a great [ecosystem](https://github.com/Level/levelup/wiki/Modules) around levelup".

> I didn't even think to look for a wiki

There's so much more :)

> I personally would rather see additional sections added to the readme over more info in the wiki

For the levelup API and encodings, yes. For userland modules, we have the wiki. That's a clear and easy  separation. It doesn't have to be a wiki per se, it can be a `docs` or `recipes` folder for all I care. But then.. a by now standard discussion will ensue between those who need to work offline and those that want small npm packages.

> I agree that the built in encoding are sparsely documented, but I do not see that as a reason to not start now.

It isn't. I just want to be careful. What do other maintainers think?

> I might decide to switch to storing protocol buffers with a binary encoding

The levelup-compatible  [protocol-buffers](https://github.com/mafintosh/protocol-buffers#leveldb-encoding-compatibility) may interest you. Which is not on the wiki, darnit.

ralphtheninja:
We are removing encodings from levelup.

Issue #423, opened 2017-05-04
package.json Metadata in Browser Bundles


kemitchell:
I noticed that browser bundles with `levelup` include a bunch of `package.json` metadata, including the content of `README`.  Best I can tell, `package.json` ends up in the `require` graph from `lib/util.js`, which loads both `levelup`'s and `leveldown`'s `package.json` files for a compatibility check.  When installing `levelup` with `npm`, the CLI pulls `package.json` down from the registry.  Registry package metadata objects inline lots of additional data, like `README` content.

Would y'all welcome a patch that adds `prepublish` and `pretest` scripts to extract versions from `levelup` and `leveldown` `package.json` into standalone files, and has `lib/util.js` require those files, instead?

juliangruber:
SGTM!

PR #424, opened 2017-05-05
Pull LevelDOWN loader out to non-Browserified module

kemitchell:
Following on from #423.

This patch pulls the LevelDOWN loading and version checking code out of `lib/util.js` into its own, separate module.  It also modifies `package.json` to prevent Browserify from packing the new module, which in turn prevents it from packing the project's `package.json` file.

I mentioned a different approach in #423: Using a prepublish lifecycle script to extract `devDependencies.leveldown` into a one-line JSON file for the version checking code to `require`.  This modular approach seems cleaner, and leverages mechanisms, like the `browser` field, that `levelup` is already using.

### DIFF
--- a/lib/levelup.js
+++ b/lib/levelup.js
@@ -26,7 +26,7 @@ var EventEmitter        = require('events').EventEmitter
 
   , getOptions          = util.getOptions
   , defaultOptions      = util.defaultOptions
-  , getLevelDOWN        = util.getLevelDOWN
+  , getLevelDOWN        = require('./leveldown')
   , dispatchError       = util.dispatchError
   , isDefined           = util.isDefined
 
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "tape": "~4.2.1"
   },
   "browser": {
+    "./lib/leveldown.js": false,
     "leveldown": false,
     "leveldown/package": false,
     "semver": false
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
   },
   "devDependencies": {
     "async": "~1.5.0",
+    "browserify": "^14.3.0",
     "bustermove": "~1.0.0",
     "tap": "~2.3.1",
     "delayed": "~1.0.1",
--- a/test/browserify-test.js
+++ b/test/browserify-test.js
@@ -1,0 +1,29 @@
+/* Copyright (c) 2012-2016 LevelUP contributors
+ * See list at <https://github.com/level/levelup#contributing>
+ * MIT License <https://github.com/level/levelup/blob/master/LICENSE.md>
+ */
+
+var common  = require('./common')
+
+  , assert  = require('referee').assert
+  , refute  = require('referee').refute
+  , buster  = require('bustermove')
+  , browserify = require('browserify')
+  , path = require('path')
+
+var PACKAGE_JSON = path.join(__dirname, '..', 'package.json')
+
+buster.testCase('Browserify Bundle', {
+  'does not contain package.json': function (done) {
+    var b = browserify(path.join(__dirname, '..'), {browserField: true})
+      .once('error', function (error) {
+        assert.fail(error)
+        done()
+      })
+    b.pipeline
+      .on('file', function (file, id, parent) {
+        refute.equals(file, PACKAGE_JSON)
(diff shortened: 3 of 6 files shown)


juliangruber:
This is even better, I very much +1!

lib/levelup.js
```diff
@@ -26,7 +26,7 @@ var EventEmitter        = require('events').EventEmitter
 
   , getOptions          = util.getOptions
   , defaultOptions      = util.defaultOptions
-  , getLevelDOWN        = util.getLevelDOWN
+  , getLevelDOWN        = require('./leveldown')
```
ralphtheninja: So this is the seam where browserify will not package `getLevelDOWN()`?

juliangruber: yep, it will return `null` or something falsy

ralphtheninja: :shipit:

juliangruber:
@ralphtheninja does that mean you agree to merge? :)

juliangruber:
oh, and what about adding a test for this one? could use `browserify --list` or something (or its JS api) to verify the package.json isn't in the bundle

ralphtheninja:
> @ralphtheninja does that mean you agree to merge? :)

Yep! colonshipitcolon :D

kemitchell:
It's my patch, but I'm not fond of this "agreement to merge" concept.

juliangruber:
> It's my patch, but I'm not fond of this "agreement to merge" concept.

can you clarify please? I was thinking another +1 would be great to have.

juliangruber:
thank you for adding the test! the logic looks great to me, just needs to be cleaned up and formatted a little

kemitchell:
I don't think the test is working quite yet.  It's seeing a `file` event for `package.json`.

I'll have another look.  This is my first time using Browserify's Node API directly like this.

kemitchell:
Needed relative paths in the browser field.  I went ahead and added both `package.json` and `lib/leveldown.js`, which `require`s it, just to be sure.

kemitchell:
Git weirdness. Will correct. Don't merge yet!

kemitchell:
ROFL. Not sure how I made myself committer of 26 prior commits.  Gonna have to check my shell history...

Anyway, this should now be ready to land.  Might not hurt to have someone who knows Browserify's Node  API better take a look at the test code, though.

package.json
```diff
@@ -57,6 +58,8 @@
     "tape": "~4.2.1"
   },
   "browser": {
+    "./package.json": false,
```
juliangruber: ah, with `./lib/lveldown.js` included, can we omit this line?

test/browserify-test.js
```diff
@@ -0,0 +1,29 @@
...
+  , refute  = require('referee').refute
+  , buster  = require('bustermove')
+  , browserify = require('browserify')
+  , path = require('path')
+
+var PACKAGE_JSON = path.resolve(path.join(__dirname, '..', 'package.json'))
+
+buster.testCase('Browserify Bundle', {
+  'does not contain package.json': function (done) {
+    var b = browserify('.', {browserField: true})
```
juliangruber: i'd prefer `browserify(path.join(__dirname, '..'), {browserField: true})` so this test works independent of your current working directory

test/browserify-test.js
```diff
@@ -0,0 +1,29 @@
...
+
+var common  = require('./common')
+
+  , assert  = require('referee').assert
+  , refute  = require('referee').refute
+  , buster  = require('bustermove')
+  , browserify = require('browserify')
+  , path = require('path')
+
+var PACKAGE_JSON = path.resolve(path.join(__dirname, '..', 'package.json'))
```
juliangruber: I think there's no need for the `path.resolve()` here, `path.join` with `__dirname` will give you an absolute path already

kemitchell:
All nits picked. Think we're green to go.

juliangruber:
awesome, thank you so much! @ralphtheninja new patch release?

ralphtheninja:
> awesome, thank you so much! @ralphtheninja new patch release?

SGTM!

juliangruber:
1.3.6 it is 🎆

Issue #426, opened 2017-05-24
getLevelDown is not a function.


diasdavid:
With 1.3.6 release, I've started getting errors of `getLevelDown is not a function` coming from a browserified version of this module and I've narrowed down the problem to the usage at:

- https://github.com/Level/levelup/blob/master/lib/levelup.js#L114
- https://github.com/Level/levelup/blob/master/lib/levelup.js#L385

The function should only be called if it actually exists, i.e `getLevelDOWN && getLevelDOWN()` or sorts. Would you take a PR with this? This change is giving me a hard time to circumvent since it was released as a patch version.

It would also be great if this module got its tests running on the browser.

ralphtheninja:
Oh right. So apparently browserifying the code does _not_ make requiring `lib/leveldown.js` something falsey but instead an empty object. cc @juliangruber

diasdavid:
All good now. Thank you so much for responding so quickly! ❤️

PR #427, opened 2017-05-24
fix: avoid calling getLevelDOWN if not present

diasdavid:
Hi there!

This is the quick fix for the issue described on: https://github.com/Level/levelup/issues/426

I intended to come up with a full PR that would run tests in the browser as well, but due to the quantity and the fact that they rely on `fs`, I haven't been able to finish this quickly, you can see progress at https://github.com/diasdavid/levelup/tree/test/browser

Here is what I propose: Merge this PR and do a patch release so that no user hits this breakage when using it in the browser. After, finish browser testing so that this situation doesn't happen again.

### DIFF
--- a/lib/levelup.js
+++ b/lib/levelup.js
@@ -30,6 +30,10 @@ var EventEmitter        = require('events').EventEmitter
   , dispatchError       = util.dispatchError
   , isDefined           = util.isDefined
 
+if (typeof getLevelDOWN !== 'function') {
+  getLevelDOWN = function noop () {}
+}
+
 function getCallback (options, callback) {
   return typeof options == 'function' ? options : callback
 }


ralphtheninja:
:+1: @diasdavid Ok, if we merge the standard PR first?

gediminasml:
@diasdavid Are you guys sure about this solving the issue? For some reason I see that `getLevelDOWN` is just an empty object `{}` in the debugger . In this case the check won't help and it'll try calling the empty object as a function.  At least it's the case when I'm debugging the package when running IPFS node in a browser.

lib/levelup.js
```diff
@@ -111,7 +111,7 @@ LevelUP.prototype.open = function (callback) {
 
   this._status = 'opening'
   this.db      = new DeferredLevelDOWN(this.location)
-  dbFactory    = this.options.db || getLevelDOWN()
+  dbFactory    = this.options.db || getLevelDOWN && getLevelDOWN()
```
ralphtheninja: We should probably make a `typeof getLevelDOWN === 'function'` check instead

juliangruber: +1

gediminasml: 👍

lib/levelup.js
```diff
@@ -30,6 +30,10 @@ var EventEmitter        = require('events').EventEmitter
   , dispatchError       = util.dispatchError
   , isDefined           = util.isDefined
 
+if (typeof getLevelDOWN !== 'function') {
+  getLevelDOWN = function noop () {}
+}
```
diasdavid: @ralphtheninja made it more simple as suggested :)

ralphtheninja:
LGTM

ralphtheninja:
1.3.7

diasdavid:
Ok, almost there, now I'm getting: `TypeError: dbFactory is not a function`. We just need to make sure that is also a function. Wanna do that in your Standard PR @ralphtheninja, avoiding an extra rebase?

ralphtheninja:
@diasdavid I haven't rebased yet, I'll wait with that until this is settled.

ralphtheninja:
@juliangruber Hilfe! :)

Issue #416, opened 2017-01-26
move to standard


juliangruber:
Shall we do this? Easier contribution and no worries about style changes.

diasdavid:
yes please! :)

ralphtheninja:
@juliangruber At time of writing you have 8 thumbs up and 2 hearts and no negative reaction. *nudge*nudge*

juliangruber:
i'm too lazy rn, does anyone else wanna do this? :P

ralphtheninja:
I can have a look

ralphtheninja:
@juliangruber What about the other branches, e.g. `encoding-down`, `remove-encodings` etc? Should I just go ahead and convert `master` and worry about the other branches later?

juliangruber:
i think so, yes. we can move them later

PR #428, opened 2017-05-24
add standard

ralphtheninja:
Closes https://github.com/Level/levelup/issues/416


test/benchmarks/tests/get_int_string_x10_sqlite3.js
```diff
@@ -1,2 +1,2 @@
 module.exports = require('./get_int_string_x1000_sqlite3').fn.bind(null, 10)
-module.exports.setup = require('./get_int_string_x1000_sqlite3').setupFn.bind(null, 10)
\ No newline at end of file
+module.exports.setup = require('./get_int_string_x1000_sqlite3').setupFn.bind(null, 10)
```
ralphtheninja: As I was working on converting I noticed that the benchmarks for sqlite3 are commented out. Maybe we should clean up these benchmarks completely?

test/common.js
```diff
@@ -4,40 +4,40 @@
...
-  , assertMessage: '${0} expected to be instance of ${1}'
-  , refuteMessage: '${0} expected not to be instance of ${1}'
+  assert: function (actual, expected) {
+    return actual instanceof expected
+  },
+  refute: function (actual, expected) {
+    return !(actual instanceof expected)
+  },
+  assertMessage: '${0} expected to be instance of ${1}', // eslint-disable-line
+  refuteMessage: '${0} expected not to be instance of ${1}' // eslint-disable-line
```
ralphtheninja: Had to add this exception since standard thinks we are trying to use a es6 template string, which we aren't.

test/common.js
```diff
@@ -4,40 +4,40 @@
...
-  , assertMessage: '${0} expected to be undefined'
-  , refuteMessage: '${0} expected not to be undefined'
+  assert: function (actual) {
+    return actual === undefined
+  },
+  refute: function (actual) {
+    return actual !== undefined
+  },
+  assertMessage: '${0} expected to be undefined', // eslint-disable-line
+  refuteMessage: '${0} expected not to be undefined' // eslint-disable-line
```
ralphtheninja: Ditto

test/common.js
```diff
@@ -101,7 +95,7 @@ module.exports.loadBinaryTestData = function (callback) {
 module.exports.binaryTestDataMD5Sum = '920725ef1a3b32af40ccd0b78f4a62fd'
 
 module.exports.checkBinaryTestData = function (testData, callback) {
-  var md5sum = crypto.createHash('md5');
+  var md5sum = crypto.createHash('md5')
```
ralphtheninja: OMG we had a semicolon in there!!11oneone

test/encoding-test.js
```diff
@@ -4,112 +4,104 @@
...
+      })
+    })
+  },
+  'test batch op encoding': function (done) {
+    this.openTestDatabase({ valueEncoding: 'json' }, function (db) {
+      db.batch([
+        {
+          type: 'put',
+          key: new Buffer([1, 2, 3]), // eslint-disable-line
+          value: new Buffer([4, 5, 6]), // eslint-disable-line
```
ralphtheninja: Had to add this exception as well since standard complains on not using `Buffer.from()` since want to test this on node 0.12 as well, which doesn't have `Buffer.from()`.

test/benchmarks/tests/get_int_string_x10_sqlite3.js
juliangruber: +1

test/common.js
juliangruber: :D

test/encoding-test.js
vweevers: How about a more specific `// eslint-disable-line node/no-deprecated-api`?

diasdavid: A better option is to use `safe-buffer` module http://npmjs.org/safe-buffer

juliangruber: this is just in the tests, so it's not super relevant

ralphtheninja: Good idea. I'll update!

ralphtheninja: > this is just in the tests, so it's not super relevant

Agreed. But a lot cleaner than weird eslint-comments here and there :)

ralphtheninja:
@juliangruber I've updated this branch based on the latest release

package.json
```diff
@@ -66,7 +68,8 @@
     "semver": false
   },
   "scripts": {
-    "test": "tape test/*-test.js | faucet"
+    "standard": "standard",
+    "test": "npm run standard && tape test/*-test.js | faucet"
```
juliangruber: why the double standard task? what about just `standard & & tape ...`

ralphtheninja: No reason :) Updating.

ralphtheninja:
No one seems to complain. Lets do this.

ralphtheninja:
@juliangruber Maybe we should run the old tests (before converting to standard) on the updated code before we release a new version

juliangruber:
i would say let's not release this at all, it's just a style change and so shouldn't concern the public. or we could release a patch, because the READMEs have updated

PR #429, opened 2017-05-24
fix: return noopnoop

diasdavid:
Started getting ` TypeError: dbFactory is not a function` now because I missed understanding that getLevelDOWN should return a function.

It is a bit weird to me that the module requires calling that function in the first place if no options are passed nor levelDOWN exists.


diasdavid:
Nvm, apparently the last issue was that `Object.assign` no longer works as normally when Webpacked. `Object.assign({}, a, b, c, ..)` will ignore `b, c, ...`, this was passing an undefined value on the db key of options object..

PR #430, opened 2017-05-24
throw more descriptive error if db factory is not a function

ralphtheninja:
`options.db` _must_ be set when `levelup` is browserified, this patch makes it clearer if/when this happens

* undos https://github.com/Level/levelup/pull/427
* related https://github.com/Level/levelup/pull/424

### DIFF
--- a/lib/levelup.js
+++ b/lib/levelup.js
@@ -30,10 +30,6 @@ var EventEmitter        = require('events').EventEmitter
   , dispatchError       = util.dispatchError
   , isDefined           = util.isDefined
 
-if (typeof getLevelDOWN !== 'function') {
-  getLevelDOWN = function noop () {}
-}
-
 function getCallback (options, callback) {
   return typeof options == 'function' ? options : callback
 }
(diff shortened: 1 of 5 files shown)


ralphtheninja:
I'll rebase the standard branch on top of this, once we have 1.3.8 out.

lib/levelup.js
```diff
@@ -116,7 +112,12 @@ LevelUP.prototype.open = function (callback) {
   this._status = 'opening'
   this.db      = new DeferredLevelDOWN(this.location)
   dbFactory    = this.options.db || getLevelDOWN()
```
juliangruber: this line will throw when you're in the browser and `this.options.db` isn't set, as `getLevelDown === {}`

juliangruber:
we should _definitely_ add tests for this

lib/levelup.js
ralphtheninja: Yes you're right. So this PR is basically useless :)

ralphtheninja: Hmm, we could wrap this and throw a custom error, e.g.

```js
try {
  dbFactory = this.options.db || getLevelDOWN()
} catch (e) {
  throw new Error('missing db factory, you need to set options.db')
}
```

And add a test for it as well.

juliangruber: sgtm

ralphtheninja:
@juliangruber I've updated the PR and added some tests for not working and working browserify code

ralphtheninja:
1.3.8

Will rebase standard branch on top of this :)

Issue #438, opened 2017-07-26
pass in leveldown instance instead of location


juliangruber:
```js
var db = levelup(leveldown(location))
```

ralphtheninja:
I guess this means all code related to `location` needs to die.

juliangruber:
yes! all `location` logic needs to go, and the lazy open feature needs to be factored in

Issue #443, opened 2017-07-29
use .status property from deferred-leveldown@2.0.0


ralphtheninja:
(note, `.status` really is from `abstract-leveldown`)

ralphtheninja:
Add test for this, see

![validate-down](https://user-images.githubusercontent.com/308049/28802469-ddaa88d4-7656-11e7-9b89-a4f21d5dc862.png)

PR #446, opened 2017-08-01
Leveldown instance as parameter

ralphtheninja:
Closes https://github.com/Level/levelup/issues/438 https://github.com/Level/levelup/issues/443


lib/levelup.js
```diff
@@ -62,14 +61,17 @@ function LevelUP (location, options, callback) {
...
 
-  if ((!options || typeof options.db !== 'function') && typeof location !== 'string') {
-    error = new InitializationError(
-        'Must provide a location for the database')
-    if (callback) {
-      return process.nextTick(function () {
-        callback(error)
-      })
-    }
+  if (!options || typeof options.db !== 'function') {
```
ralphtheninja: @juliangruber If we're happy with this PR we might want to consider squashing this commit with the next. It was easier for me to follow along what you did on `v2` branch before removing `location` completely.

package.json
```diff
@@ -61,10 +60,8 @@
     "tape": "^4.7.0"
   },
   "browser": {
-    "./lib/leveldown.js": false,
     "leveldown": false,
-    "leveldown/package": false,
-    "semver": false
+    "leveldown/package": false
```
ralphtheninja: Are these really needed now? It seems to me we can remove the whole `browser` object unless there's something I'm missing.

ralphtheninja:
@juliangruber If we're happy with this PR we might want to consider squashing the first and second commits into one. It was easier for me to follow along what you did on v2 branch before removing location completely, but most of the first commit is overwritten by the second one.

package.json
juliangruber: yeah! seems like we can just remove the whole thing :)

juliangruber:
aw yes!!!!

PR #455, opened 2017-08-25
Promisify

juliangruber:
This is a first stab at implementing a promise api for core. The convention is as expected: When you don't pass a callback, a promise will be returned. We're using native promises exclusively.

The library used to promisify the api lives in `lib/util`, until we decide if this is the way to go. This approach of setting a new `callback()` was the simplest least invasive way I could think of. Happy to find an easier solution!

For those who are concerned in keeping the performance of the callback api: The promise logic will _only_ be used when you don't pass a callback.

The _breaking_ change here is: If you call `db.put('foo', 'bar')` without a callback, as sometimes seen in example code, it will delegate error handling to the returned promise instead of emitting an `"error"` event on the `db` object.

What do you all think?

cc @pfrazee @madbence @limoragni @fczuardi

TODO

- [x] <del>fix on node 4</del> deprecate node 0.12 and 4
- [x] add tests
- [x] add docs


juliangruber:
@ralphtheninja Tests are failing on node <6. Are we keeping node 4 and 0.12?

pfrazee:
Looks good to me (at a glance)

juliangruber:
Question, that is very likely to get a 100% no, but still worth being agreed on: Should we drop the callback interface completely?

[](https://m131jyck4m.execute-api.us-west-2.amazonaws.com/prod/poll/01BRDE5C3YHP41D037S58MD7JR/Drop%20callbacks/vote)
[](https://m131jyck4m.execute-api.us-west-2.amazonaws.com/prod/poll/01BRDE5C3YHP41D037S58MD7JR/Keep%20callbacks/vote)

ralphtheninja:
> Happy to find an easier solution!

Creating the promise is a oneliner. Hard to find an easier solution :)

ralphtheninja:
> @ralphtheninja Tests are failing on node <6. Are we keeping node 4 and 0.12? (Because of [ foo, bar ] = ret() assignment)

I personally wouldn't mind dropping 4 and 0.12.

We should also use `const`.

ralphtheninja:
Oh. And docs as well. But we could push that to a separate issue.

ralphtheninja:
@juliangruber How did you create that poll? Was really neat :)

juliangruber:
> @juliangruber How did you create that poll? Was really neat :)

That was https://github.com/apex/gh-polls :)

juliangruber:
let's do the es6-ification in another PR

lib/util.js
```diff
@@ -23,9 +23,22 @@ function isDefined (v) {
...
+  var resolve_, reject_
+  var promise = new Promise(function (resolve, reject) {
+    [resolve_, reject_] = [resolve, reject]
+  })
+  function callback (err, value) {
+    if (err) reject_(err)
+    else resolve_(value)
+  }
+  return [ callback, promise ]
+}
```
vweevers: How about:

```js
function promisify () {
  var callback
  var promise = new Promise(function (resolve, reject) {
    callback = function callback (err, value) {
      if (err) reject(err)
      else resolve(value)
    }
  })
  
  return [ callback, promise ]
}
```

juliangruber: great idea :)

juliangruber:
For those who are concerned in keeping the performance of the callback api: The promise logic will _only_ be used when you don't pass a callback.

juliangruber:
added tests and docs :) can we get reviews on those too?

README.md
```diff
@@ -189,7 +194,7 @@ Encoding of the `key` and `value` objects will adhere to `'keyEncoding'` and `'v
 --------------------------------------------------------
 <a name="get"></a>
 ### db.get(key[, options][, callback])
-<code>get()</code> is the primary method for fetching data from the store. The `key` can be an arbitrary data object. If it doesn't exist in the store then the callback will receive an error as its first argument. A not-found err object will be of type `'NotFoundError'` so you can `err.type == 'NotFoundError'` or you can perform a truthy test on the property `err.notFound`.
+<code>get()</code> is the primary method for fetching data from the store. The `key` can be an arbitrary data object. If it doesn't exist in the store then the callback or promise will receive an error as its first argument. A not-found err object will be of type `'NotFoundError'` so you can `err.type == 'NotFoundError'` or you can perform a truthy test on the property `err.notFound`.
```
vweevers: Reads as "[the] promise will receive an error as its first argument". Perhaps simply say: "If it doesn't exist then the callback or promise will receive an error."

README.md
```diff
@@ -168,19 +169,23 @@ Additionally, each of the main interface methods accept an optional options obje
 
 However, it is possible to *reopen* a database after it has been closed with <a href="#close"><code>close()</code></a>, although this is not generally advised.
 
+If no callback is passed, a promise is returned.
```
vweevers: Maybe: "native promise". Or mention that in the Promise Support section.

README.md
```diff
@@ -427,6 +439,43 @@ Check out the implementations that the community has already produced [here](htt
...
+
+Each function taking a callback also can be used as a promise, if the callback is omitted. This applies for:
+
+- `db.get(key[, options])`
+- `db.put(key, value[, options])`
+- `db.del(key[, options])`
+- `db.batch(ops[, options])`
+- `db.batch().write()`
+
+The only exception is the `levelup` constructor itself, which if no callback is passed will lazily open the database backend in the background.
```
vweevers: Nitpick (and a pre-existing problem): throughout the readme we're mixing "database" /  "database store" / "store" / "database backend" / "LevelDB store".

README.md
```diff
@@ -427,6 +439,43 @@ Check out the implementations that the community has already produced [here](htt
...
+
+Example:
+
+```js
+var db = levelup(leveldown('./my-db'))
+
+db.put('foo', 'bar')
+  .then(function () { return db.get('foo') })
+  .then(function (value) { console.log(value) })
```
vweevers: Let's add a `catch` handler too, for completeness and advocating best practices :)

README.md
juliangruber: addressed

README.md
juliangruber: addressed

README.md
juliangruber: addressed

README.md
ralphtheninja: I think we should spend some time on the readme in general. It seems to me that it has too much information as well as outdated

juliangruber: I agree! This should live in its own Issue/PR though

lib/util.js
ralphtheninja: I'm curious how this _actually_ works. Is it because of `function callback () {}`? `promisify()` returns `[ undefined, new Promise() ]` right? And once `promise.then()` is called by the caller, then the Promise-callback sets `callback` to `function callback () {}`.

vweevers: @ralphtheninja it works because the `function (resolve, reject) {}` is called immediately by the Promise constructor, before the constructor even returns.

ralphtheninja: @vweevers Thanks for the lesson! :) I always thought that `.then()` etc actually triggered the promise. TIL.

README.md
```diff
@@ -427,6 +439,43 @@ Check out the implementations that the community has already produced [here](htt
...
+
+Or using `async/await`:
+
+```js
+const main = async () {
+  const db = levelup(leveldown('./my-db'))
+
+  await db.put('foo', 'bar')
+  console.log(await db.get('foo'))
+}
```
ralphtheninja: :+1:

ralphtheninja:
LGTM on docs and tests etc. Great work @juliangruber (as always)!!

README.md
vweevers: See #456.

juliangruber:
awesome! thank you for the reviews :) merging then into the 2.x branch

rvagg:
So, just to clarify, we explicitly dropped v4.x support with this PR, was there explicit agreement on that? I only see a couple of casual references here. Note that Node 4 is still supported upstream and widely used https://nodejs.org/metrics/summaries/version.png

I'm not objecting, I'm not active enough to make such an objection, just wanted to hear that it was a conscious group decision.

juliangruber:
It didn't seem like a big deal as there was no objection from either of the maintainers, so no discussion arose. We shipped this as a major, so it was ok to break things in `2.x`. Going forward however we can discuss whether we want to go full on modern JS or add back older node support via for example a promise shim.

juliangruber:
Also, very good to hear from you again, @rvagg :)

MeirionHughes:
I was under the impression the idea was to drop support when node x left maintenance: https://github.com/Level/levelup/issues/470

PR #458, opened 2017-08-29
Remove encodings, take two

ralphtheninja:

PR #461, opened 2017-08-29
Maybe error sync

ralphtheninja:
`callback` is called within `maybeError()` before we even return the `promise`


ralphtheninja:
Should be enough to merge this to close https://github.com/Level/levelup/pull/458

juliangruber:
have you tested this? i think it won't work because when we do `[callback,promise] = promisify()` we create a new `callback` reference, different from the one passed to `maybeError`. This means all calls to `maybeError` should go after creating the promise, right?

ralphtheninja:
> have you tested this? i think it won't work because when we do [callback,promise] = promisify() we create a new callback reference, different from the one passed to maybeError. This means all calls to maybeError should go after creating the promise, right?

Nope. Hold off with this PR and let me write a test for it first. I might be wrong here.

ralphtheninja:
Closing. Will make a new PR if I'm right :wink:

PR #462, opened 2017-08-29
maybeError() calls callback synchronously

ralphtheninja:
Seems I was correct.

### DIFF
--- a/test/.gitignore
+++ b/test/.gitignore
@@ -1,0 +1,1 @@
+_levelup_test_db_61
--- a/test/.gitignore
+++ b/test/.gitignore
@@ -1,1 +1,1 @@
-_levelup_test_db_61
+_levelup_test_db*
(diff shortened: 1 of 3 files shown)


test/.gitignore
```diff
@@ -0,0 +1 @@
+_levelup_test_db_61
```
ralphtheninja: Doh, did `echo _levelup_test_db* >> .gitignore`

juliangruber:
ah, yes, but this won't work with the promises

ralphtheninja:
> ah, yes, but this won't work with the promises

Why not?

ralphtheninja:
@juliangruber I don't think promises work if we _don't_ do this. That is, the opposite of what your stating. But again, I might be wrong.

vweevers:
And I think *both* would work 😄  Under the assumption that it doesn't matter when the callback is called; the promise will always resolve (to its consumer) in the next tick.

ralphtheninja:
As long as I'm not messing up promises I'm happy. I can add a few similar tests using promises if you like.

juliangruber:
I should have looked at the code! :D

It actually doesn't matter for promises, only for the regular callback case. With promises, you can resolve them before or after you return them.

+1 then!

PR #469, opened 2017-09-02
readme: add node version badge

ralphtheninja:


package.json
```diff
@@ -61,5 +61,8 @@
   "scripts": {
     "test": "standard && tape test/*-test.js | faucet"
   },
+  "engines": {
+    "node": ">=6.0.0"
```
vweevers: not node 5?

README.md
```diff
@@ -5,6 +5,7 @@ LevelUP
 
 > Fast and simple storage. A node.js wrapper for `abstract-leveldown` compliant backends.
 
+![Node version](https://img.shields.io/badge/Node.js-%3E%3D6.0.0-orange.svg?style=flat-square)
```
vweevers: With `engines.node` set, you can use `Node version` (badge becomes available after npm publish)

package.json
ralphtheninja: Afaik destructuring doesn't work with node 5

README.md
ralphtheninja: Aha, didn't know this. I was wondering why `leveldown` had a "n/a" badge.

ralphtheninja: This is sort of bad for native modules (here is ok I guess) since we don't want to publish a new `leveldown` version just to get a working badge in the readme.

package.json
ralphtheninja: We should rewrite the `[ callback, promise ] = promisify()` logic in this case

README.md
vweevers: Hmm yeah, so let's switch 'em :) have a hardcoded badge in leveldown until the next release

ralphtheninja: Fixed.

package.json
vweevers: I don't feel strongly about it. If someone needs node 5 back they can do a PR.

But it does mean we're no longer supporting node versions until their end of life.

ralphtheninja: Then I think we should put back node 5 on travis and fix the code to support it.

Added it back here to test it out https://travis-ci.org/ralphtheninja/levelup/jobs/271278902

ralphtheninja: We don't really need to use destructuring here, might as well just return an object with `.callback` and `.promise` set.

ralphtheninja:
> I don't feel strongly about it. If someone needs node 5 back they can do a PR.

Or we wait, as you say and if there are people complaining we can fix it easily.

vweevers:
Sure. I don't want to get in the way of your amazing hard work :) And in any case it's fine for an `rc`.

ralphtheninja:
> Sure. I don't want to get in the way of your amazing hard work :) And in any case it's fine for an rc.

You're not in the way. Always give awesome feedback!

PR #479, opened 2017-09-10
Tweak ts tests

ralphtheninja:
Turns out `faucet` is causing problems. Also don't need buster to run `ts-node`.


ralphtheninja:
Supersedes https://github.com/Level/levelup/pull/475

MeirionHughes:
Yeah, I was going to make the same change; but had to pop out quickly. 

Mind if I pull from this fork to carry on? I still think there are a few refinements to be made before merger.

ralphtheninja:
@MeirionHughes Sure go ahead!

vweevers:
Instead of spawning a ts-node process for each test file, I suggest:

```json
"scripts": {
    "test": "standard && node test | faucet && npm run ts-test",
    "ts-test": "ts-node --no-cache test"
}
```

Then in `test/index.js`:

```js
const fs = require('fs')
const path = require('path')

fs.readdirSync(__dirname)
  .filter(f => /-test.js/.test(f))
  .forEach(f => require(path.join(__dirname, f)))
```

That's basically what tape was doing anyway, and this works for both node and ts-node. WDYT?

Also when I tested this, I had to do `ts-node --no-cache` to get it to pick up changes to the type definitions. Any harm to keeping that flag?

ralphtheninja:
@vweevers PR! :wink:

vweevers:
Against which branch?

ralphtheninja:
@vweevers master

MeirionHughes:
I've just pulled from upstream and I think this broke the test for me; 

```
> levelup@2.0.0-rc1 test D:\Code\levelup
> standard && tape test/*-test.js | faucet && npm run ts-test

module.js:491
    throw err;
    ^

Error: Cannot find module 'D:\Code\levelup\test\*-test.js'
... 
```

vweevers:
@MeirionHughes can you try my branch (#480)?

MeirionHughes:
just about to pull from it. :P

I wish github pr requests would give you the pull url for a pr. :D

MeirionHughes:
ahh, lol, you didn't fork. I was trying to add you as a remote. doh!

MeirionHughes:
@vweevers yes, that appears to have fixed it - its running on Win10

PR #481, opened 2017-09-11
Improve Typings

MeirionHughes:
use AbstractLevelDOWN typings and infer types from underlying db. 

* [x] leveldown typings
* [x] memdown typings
* [x] rocksdb typings
* [x] leveldown-hyper typings
* [x] add note on typings in readme


MeirionHughes:
@ralphtheninja hold off for a bit. Need to check something with the StreamOptions.

MeirionHughes:
@ralphtheninja Sorry, for all the refactoring going on. These typings are just very complicated, but ultimately should make the usage for the end-user very simple and all the generics will be inferred automatically. 

I've added (/sigh) another generic for the `TIteratorOptions` so the abstract-leveldown implementations can define what they need themselves. 

However,  because [the docs specify the iterator options in createReadStream](https://github.com/Level/levelup#dbcreatereadstreamoptions) I've kept [ReadStreamOptions](https://github.com/MeirionHughes/levelup/blob/5a0a3eee00bcf33afffd9ed8a8f46284dbc17e75/lib/levelup.d.ts#L63-L75) for now and [just merge them with TIteratorOptions](https://github.com/MeirionHughes/levelup/blob/5a0a3eee00bcf33afffd9ed8a8f46284dbc17e75/lib/levelup.d.ts#L41-L43)

ralphtheninja:
> @ralphtheninja Sorry, for all the refactoring going on. These typings are just very complicated, but ultimately should make the usage for the end-user very simple and all the generics will be inferred automatically.

No worries. No contributor should be sorry for contributing! :wink:

MeirionHughes:
... it actually works. (I added types to my local leveldown)

image

😌

ralphtheninja:
@MeirionHughes So I guess we need to pull in `abstract-leveldown@2.7.0` separately as a dev dependency in order for this to work, right?

MeirionHughes:
@ralphtheninja  you could do either I think; but given the module is now directly used during dev; a devDependency probably makes sense. 

Also, I've finally hit a limitation of typescript with all the generics. 

if you do: 
```ts
let db = levelup<string, number>(leveldown('./db'));
``` 
then you lose the inference for the remaining generics, and they default to `any`.

ralphtheninja:
@MeirionHughes something's up with travis. I've tried locally with latest leveldown and memdown (greenkeeper branches with abstract-leveldown@2.7.0)

MeirionHughes:
@ralphtheninja looks like the custom encoder typing is too strict. its this that is failing

https://github.com/Level/levelup/blob/master/test/json-test.js#L22-L24

[I read this](https://github.com/Level/codec#encodings) and assumed they were all required. I presume they're all optional? or should a custom encoder have all 4 fields?

ralphtheninja:
> I read this and assumed they were all required. I presume they're all optional? or should a custom encoder have all 4 fields?

Iirc they are all required, see https://github.com/Level/codec/blob/master/lib/encodings.js

@juliangruber Please correct me if I'm wrong.

MeirionHughes:
btw, if I change the test to be: 
```js
valueEncoding: {
  encode: JSON.stringify,
  decode: JSON.parse,
  type: "json",
  buffer: false
}
```

then typescript is happy.

ralphtheninja:
@MeirionHughes Oh I think we talk about different things. In `level/codec` the different _encodings_ have four required properties. But when you specify an encoding for keys and/or values, you can omit the encoding for e.g. keys, but have a specific encoding for values only.

Does that make sense?

ralphtheninja:
> btw, if I change the test to be:
> 
> valueEncoding: {
>   encode: JSON.stringify,
>   decode: JSON.parse,
>   type: "json",
>   buffer: false
> }
> 
> then typescript is happy.

Aah good catch. We should update the test then I think.

ralphtheninja:
@MeirionHughes Or rather just use `{ valueEncoding: 'json' }` since we have already specified that here https://github.com/Level/codec/blob/master/lib/encodings.js#L16-L21

test/json-test.js
```diff
@@ -21,7 +21,9 @@ buster.testCase('JSON API', {
         levelup(encDown(leveldown(location), {
           valueEncoding: {
             encode: JSON.stringify,
-            decode: JSON.parse
+            decode: JSON.parse,
+            type: "json",
+            buffer: false
```
ralphtheninja: Does `{ valueEncoding: 'json' }` work?

ralphtheninja: It _should_ and also DRY to use https://github.com/Level/codec/blob/master/lib/encodings.js#L16-L21

MeirionHughes: yup. I will replace the commit

MeirionHughes: ah, I think I missed those lib exports from the codec typings.

ralphtheninja:
@MeirionHughes Works fine! Anything else you want to include or are working on right now that you want to get into the next rc release?

MeirionHughes:
@ralphtheninja I think it should be good; I tested it locally and it all seemed to work. I'm still working on the leveldown typings, but they're starting to work too: 

image

ralphtheninja:
@MeirionHughes Cool. Let me know when you're ready with `leveldown`. There are a few PRs we should get merged there as well before releasing a new patch, `nan` and `abstract-leveldown` updated.

ralphtheninja:
Maybe we should have some typings for `memdown` done as well before releasing `rc3`.

MeirionHughes:
Yes; the `levelup` generics will all default to `any` unless you pass through something that implements `Abstract.LevelDown`, its somewhat useless unless all the base db's are typed.

ralphtheninja:
@MeirionHughes Lets keep this PR open until they are done and then we can release `rc3`. Sound good?

MeirionHughes:
yes; doing the leveldown typings caused a few levelup errors to crop up.

MeirionHughes:
@ralphtheninja [this piece of test code throws a type error](https://github.com/Level/levelup/blob/master/test/open-patchsafe-test.js#L17) because `valueEncoding` is not an option for LevelDown. if I remove the encoding option, it passes both tests.

test/open-patchsafe-test.js
```diff
@@ -14,7 +14,7 @@ function test (fun) {
   return function (done) {
     var location = common.nextLocation()
     // 1) open database without callback, opens in worker thread
-    var db = levelup(leveldown(location), { valueEncoding: 'utf8' })
+    var db = levelup(leveldown(location)/*, { valueEncoding: 'utf8' }*/)
```
ralphtheninja: Just remove `/* .. */` instead or wrap with `encoding-down` to be explicit (apparently not needed)

ralphtheninja:
I believe

```js
levelup(encoding(leveldown('./db')), {valueEncoding:'utf8'})
```

is incorrect, it should be

```js
levelup(encoding(leveldown('./db'), {valueEncoding:'utf8'}))
```

but it's not needed, since default is `'utf8'` for `encoding-down`, iirc :)

juliangruber:
exactly, levelup doesn't know anything about encodings any more

MeirionHughes:
```ts
 levelup(encoding(leveldown('./db'), {valueEncoding:'utf8'}))
```

That's typed correctly and working fine. 

However, levelup does pass its own options to the base db in `open`. So the options type you pass to levelup(..., options),  is currently typed (for levelup->encoding->leveldown) like: 

```ts
type Options = leveldown.LevelDownOptions & encoding.CodecOptions
```

and that options object does get used and passed to the base dbs via `open`. As far as I can tell, levelup's options don't seem to be used anywhere else within levelup?

ralphtheninja:
```ts
type LevelUpOptions = leveldown.LevelDownOptions
```

Yep, that makes sense. https://github.com/Level/leveldown#leveldownopenoptions-callback

test/open-patchsafe-test.js
ralphtheninja: Also `standard` complains :wink:

MeirionHughes: cleaned up and combined commits to date.

ralphtheninja:
Gawd, travis is practically completely unusable today. It's super annoying to have to wait hours for builds to finish.

MeirionHughes:
Is there a queue on the Travis for the Level org? could rule out one of these packages causing a very long build?

package.json
```diff
@@ -43,13 +43,14 @@
...
     "after": "^0.8.2",
     "async": "^2.5.0",
     "bl": "^1.2.1",
     "browserify": "^14.3.0",
     "bustermove": "~1.0.0",
     "delayed": "~1.0.1",
-    "encoding-down": "^2.1.5",
+    "encoding-down": "^2.2.0",
     "faucet": "~0.0.1",
     "leveldown": "^1.7.2",
```
ralphtheninja: @MeirionHughes This needs to be bumped to `^1.8.0`

MeirionHughes: fixed

ralphtheninja:
@MeirionHughes Added `leveldown-hyper` to the list. Hope you don't mind.

MeirionHughes:
Sure, no worries. I'll make my way through them when I have time.

README.md
```diff
@@ -484,6 +484,14 @@ LevelDB is thread-safe but is **not** suitable for accessing with multiple proce
 
 See the <a href="https://github.com/level/levelup/wiki/Modules"><b>wiki</b></a> for some LevelUP extensions, including [multilevel](https://github.com/juliangruber/multilevel), that may help if you require a single store to be shared across processes.
 
+<a name="typings"></a>
+Typescript 
+----------
+
+LevelUP comes with typescript definitions that can automatically infer options based off a typed level-down implementation.
```
vweevers: - "based off" => "based on"?
- "level-down implementation" => we commonly stylize this as `abstract-leveldown` implementation (with backticks)

README.md
```diff
@@ -484,6 +484,14 @@ LevelDB is thread-safe but is **not** suitable for accessing with multiple proce
 
 See the <a href="https://github.com/level/levelup/wiki/Modules"><b>wiki</b></a> for some LevelUP extensions, including [multilevel](https://github.com/juliangruber/multilevel), that may help if you require a single store to be shared across processes.
 
+<a name="typings"></a>
+Typescript 
+----------
+
+LevelUP comes with typescript definitions that can automatically infer options based off a typed level-down implementation. 
+
+See the <a href="https://github.com/Level/levelup/wiki/Typings"><b>wiki</b></a> for for more information.
```
vweevers: "for for" => "for"

README.md
MeirionHughes: fixed 👍

README.md
MeirionHughes: fixed 👍

test/leveldown-substitution-test.js
```diff
@@ -7,14 +7,14 @@ var levelup = require('../lib/levelup.js')
...
-var MemDOWN = require('memdown')
+var memdown = require('memdown')
 var encDown = require('encoding-down')
 
 require('./common')
 
 buster.testCase('LevelDOWN Substitution', {
   'test substitution of LevelDOWN with MemDOWN': function (done) {
-    var md = new MemDOWN('foo')
+    var md = memdown('foo')
```
MeirionHughes: @ralphtheninja should be equivalent. I couldn't find a way to type memdown as function + newable + generic + static. It was a case of pick 3... but not all 4.

ralphtheninja:
2.0.0-rc3

ralphtheninja:
@MeirionHughes Super awesome work btw! :fireworks: :man_dancing:

juliangruber:
Awesome work, everyone!

https://twitter.com/juliangruber/status/908688876381892608

MeirionHughes:
@ralphtheninja @juliangruber 

all this just so that in typescript I can do

```ts
async getRange() {
  return {
    lower: await
      linq(this._db.iterator({ limit: 1, values: false }))
        .select(x => x[0])
        .first(),
    upper: await
      linq(this._db.iterator({ limit: 1, reverse: true, values: false }))
        .select(x => x[0])
        .first()
  }
}
```

image

🤣 

... works anyhow. 👍

PR #482, opened 2017-09-12
Add LevelUp Interface to namespace levelup

drew-y:
Allows LevelUp type annotation in typescript files.


MeirionHughes:
Hi, thanks for the feedback; The typings are currently in the process of being updated again for the next RC. I've exposed all the interfaces in the next typings: https://github.com/Level/levelup/pull/481

Issue #472, opened 2017-09-05
public levelup rc testing


juliangruber:
We should get as many people as possible to test `levelup@2.0.0-rc1`.

I'm thinking the following:

- [x] tweet about it: https://twitter.com/juliangruber/status/907246400194441217
- [ ] test libs
  - [x] ~~level-sublevel~~ subleveldown
  - [x] pouchdb
  - [ ] browserify-fs
  - [ ] merkle-patricia-tree
  - [ ] dynalite
  - [ ] mosca
  - [ ] hard-source-webpack-plugin
  - [ ] gittoken-api-middleware
  - [x] secure-scuttlebutt
- [x] test apps
  - [x] Voltra

ralphtheninja:
@juliangruber Maybe check projects on npm with rather frequent downloads that are based on levelup?

ralphtheninja:
https://www.npmjs.com/browse/depended/levelup

Maybe `pouchdb`?

juliangruber:
> @juliangruber Maybe check projects on npm with rather frequent downloads that are based on levelup?

Great idea! Didn't check downloads, but those stood out to me:

- pouchdb (dough of course)
- mosca
- levelgraph

juliangruber:
But yeah let's rather write a script that checks download numbers. Or is there an interface for this already?

juliangruber:
https://twitter.com/juliangruber/status/905149153810010112

vweevers:
https://github.com/addaleax/npm-get-top-dependents

ralphtheninja:
I think we should release an rc1 of level as well, before tweeting about it and check projects using it. In theory it's the easiest module to upgrade since it takes care of the details.

ralphtheninja:
@juliangruber You could also take a look at some of the `nice-registry` modules, e.g. https://github.com/nice-registry/download-counts

ralphtheninja:
@zeke How would you approach something like this? For example quering for modules that depend on `level` or `levelup` and maybe sort them based on download count?

zeke:
> How would you approach something like this?

```sh
git clone https://github.com/nice-registry/native-modules
cd native-modules
$EDITOR lib/check-package.js # customize for levelup
npm install
npm run collect
```

Then tweak `render.js` while the collector runs. Takes an hour or two usually.

At any time while the collector is running:

```
npm run render
```

> sort them based on download count?

[...]

PR #492, opened 2017-09-20
reverse: remove typings and typescript test

MeirionHughes:
temporarily remove typings from 2.0 `master` to unblock. typings work to continue in 2.1 branch and other repositories until stable.


MeirionHughes:
@ralphtheninja sorry for being a pain in the ass. Its getting closer though!

```ts
let down = leveldown('./db');

let encoded = encode(down, {
  keyEncoding: {
    buffer: true,
    encode: (a: Foo) => a.name,
    decode: (a) => new Foo(String(a)),
    type: "custom"
  },
  valueEncoding: {
    buffer: true,
    encode: (a: Foo) => a.name,
    decode: (a) => new Foo(String(a)),
    type: "custom"
  }
});

let up = levelup(encoded);
```

and levelup's types are automatically inferred to have `Foo` as key and value, because of the custom encoders. 

image

vweevers:
@MeirionHughes can the typings handle encodings that support multiple types? E.g.`json` or `bytewise`. Where keys and values can be strings, numbers, arrays, etc, all in the same database.

And how about sublevels with a different encoding (either using `level-sublevel` or `subleveldown`)?

MeirionHughes:
I don't see why not. if you do 
```ts
let encoded = encode(down, {
  valueEncoding:'json'
});
```
up will typed with `any`, so it will accept anything for the value (in this case the same is true for key). On the most part, unless it is explicitly typed or can be inferred from the typed encoder it will simply default to `any` for key and value types and not complain with whatever you pass through. 

I've not checked level-sublevel on my current types; but levelup typings (just the abstract + down stuff) hasn't changed at all since I last tested it: https://github.com/Level/levelup/issues/472#issuecomment-330248380

PR #494, opened 2017-09-23
add default export

zixia:
Enable us to use `import levelup from 'levelup'` by adding `default` export.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ LevelUP
   * <a href="#basic">Basic usage</a>
   * <a href="#api">API</a>
   * <a href="#promises">Promise Support</a>
+  * <a href="#import">Import with Type Definitions</a>
   * <a href="#events">Events</a>
   * <a href="#extending">Extending LevelUP</a>
   * <a href="#multiproc">Multi-process access</a>
@@ -449,6 +450,36 @@ const main = async () {
   console.log(await db.get('foo'))
 }
 ```
+
+--------------------------------------------------------
+
+<a name="import"></a>
+ES6 Import and Typing(for TypeScript)
+-------------------------------------
+
+We have two ways to import(require) the levelup module in the code.
+
+### 1. The old fashioned `require`
+
+```js
+var levelup = require('levelup')
+```
+
+### 2. The new ES6 `import`
+
+```js
+import levelup from 'levelup'
+```
+
+## TypeScript
+
+LevelUp ships with TypeScript `type definitions` support out-of-box. So we can use `import` with Type Defination by default.
+
+The benefits will be:
+1. [Static Type Checking](http://2ality.com/2014/10/typed-javascript.html)
+1. [IDE IntelliSense](https://code.visualstudio.com/docs/editor/intellisense)
+
+Recommended IDE: [Visual Studio Code](https://code.visualstudio.com/)
 
 --------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


ralphtheninja:
LGTM

juliangruber:
Should we document this as well? You too often have to guess how to `import` a node module

ralphtheninja:
@zixia Mind adding something to the README?

zixia:
@ralphtheninja Ok, I will try to add some docs for this.

zixia:
I just add some docs about to introduce the `require`/`import` and `typing`. 

Do not know where is the best to add them, so I put them after the `Promise` part.

juliangruber:
I feel this could be simpler, what about having this in the first code example:

```js
var levelup = require('levelup')
var leveldown = require('leveldown')

// or
import levelup from 'levelup'
import leveldown from 'leveldown
```

zixia:
@juliangruber No problem for me. I can modify the PR if needed.

Actually, this PR does not change any import/require usage than before.

It only affects the `rollup` tools behavior. (see my reference above)

juliangruber:
Well, it adds the ability to import the default export for any es6 modules compliant environment.

README.md
```diff
@@ -452,6 +453,36 @@ const main = async () {
 
 --------------------------------------------------------
 
+<a name="import"></a>
+ES6 Import and Typing(for TypeScript)
+-------------------------------------
+
+We have two ways to import(require) the levelup module in the code.
+
+### 1. The old fashioned `require`
```
ralphtheninja: By using `require`

README.md
```diff
@@ -452,6 +453,36 @@ const main = async () {
...
+
+```js
+var levelup = require('levelup')
+```
+
+### 2. The new ES6 `import`
```
ralphtheninja: By using ES6 `import`

README.md
```diff
@@ -452,6 +453,36 @@ const main = async () {
...
+
+```js
+import levelup from 'levelup'
+```
+
+## TypeScript
```
ralphtheninja: Maybe remove this section? Not sure what it has to do with this PR? (And we did remove typings for now until a later version)

zixia:
Edited by following reviews from @ralphtheninja

MeirionHughes:
I think typescript will have a hard time supporting _both_ with this because you're mixing module kinds: 

```ts
export = levelup;
export default levelup;
```
error:
```
[ts] An export assignment cannot be used in a module with other exported elements.
```

So to type this I'd have to pick one _kind_ for ts users.

zixia:
@MeirionHughes Yes, you can not use `export = ` with `export default` together.

We should either to use

1. in ES6/TypeScript
```ts
export { levelup }
export default levelup
```

2. in traditional
```js
export = levelup.default = levelup.levelup = levelup
```

I believe in both of above two kinds, the user could always use `import * as levelup from 'levelup` in their code.

juliangruber:
It has to stay

```js
const levelup = require('levelup')
```

for "older" systems though, is there a way to keep this _and_ make es6 environments happy?

ralphtheninja:
I'm good with the changes and agree with @juliangruber, whatever you want to pick for es6 stuff should not mess up current functionality.

zixia:
@juliangruber Yes, I agree with you that we have to stay `const levelup = require('levelup')`.

This PR did not affect the "older" systems usage because it only adds a `default` export to the library.

zixia:
When will we publish the latest version to NPM? Thanks!

MeirionHughes:
I've got the typings nearly ready to add back in to; waiting on leveldown.

PR #499, opened 2017-10-04
feat: restore typings + testing

MeirionHughes:
* to start discussion and review of adding typings back into master
* keen eye will notice imports for tests are using `.default` for some level stuff. This is because the typescript compiler will complain because those repositories now have es2015 default exports.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@
   * <a href="#events">Events</a>
   * <a href="#extending">Extending LevelUP</a>
   * <a href="#multiproc">Multi-process access</a>
+  * <a href="#typings">TypeScript </a>
   * <a href="#support">Getting support</a>
   * <a href="#contributing">Contributing</a>
   * <a href="#license">Licence &amp; copyright</a>
@@ -474,6 +475,15 @@ Multi-process access
 Stores like LevelDB are thread-safe but they are **not** suitable for accessing with multiple processes. You should only ever have a store open from a single Node.js process. Node.js clusters are made up of multiple processes so a LevelUP instance cannot be shared between them either.
 
 See the <a href="https://github.com/level/levelup/wiki/Modules"><b>wiki</b></a> for some LevelUP extensions, including [multilevel](https://github.com/juliangruber/multilevel), that may help if you require a single store to be shared across processes.
+
+<a name="typings"></a>
+TypeScript 
+----------
+
+LevelUP comes with TypeScript definitions that can automatically infer options from a typed `abstract-leveldown` implementation. 
+
+See the <a href="https://github.com/Level/levelup/wiki/Typings"><b>wiki</b></a> for  more information. 
+
 
 <a name="support"></a>
 Getting support
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "json"
   ],
   "main": "lib/levelup.js",
+  "typings": "lib/levelup.d.ts",
   "dependencies": {
     "deferred-leveldown": "~2.0.2",
     "level-errors": "~1.1.0",
@@ -41,27 +42,31 @@
     "xtend": "~4.0.0"
   },
   "devDependencies": {
-    "abstract-leveldown": "^2.7.0",
+    "@types/node": "^8.0.31",
+    "abstract-leveldown": "^2.7.1",
     "after": "^0.8.2",
     "async": "^2.5.0",
     "bl": "^1.2.1",
     "browserify": "^14.3.0",
     "bustermove": "~1.0.0",
     "delayed": "~1.0.1",
-    "encoding-down": "^2.2.0",
+    "encoding-down": "^2.3.1",
     "faucet": "~0.0.1",
     "leveldown": "^2.0.0",
-    "memdown": "^1.2.4",
+    "memdown": "^1.4.1",
     "msgpack-js": "~0.3.0",
     "referee": "~1.2.0",
     "rimraf": "^2.6.1",
     "safe-buffer": "^5.1.0",
     "slow-stream": "0.0.4",
     "standard": "^10.0.2",
-    "tape": "^4.7.0"
+    "tape": "^4.7.0",
+    "ts-node": "^3.3.0",
+    "typescript": "^2.5.2"
   },
   "scripts": {
-    "test": "standard && node test | faucet"
+    "test": "standard && node test | faucet && npm run ts-test",
+    "ts-test": "ts-node --no-cache test | faucet"
   },
   "engines": {
     "node": ">=6.0.0"
--- a/test/common.js
+++ b/test/common.js
@@ -12,11 +12,12 @@ var rimraf = require('rimraf')
 var fs = require('fs')
 var path = require('path')
 var delayed = require('delayed').delayed
-var levelup = require('../lib/levelup.js')
+
+var levelup = require('../lib/levelup.js').default
 var errors = require('level-errors')
 var dbidx = 0
-var leveldown = require('leveldown')
-var encDown = require('encoding-down')
+var leveldown = require('leveldown').default
+var encDown = require('encoding-down').default
 
 assert(levelup.errors === errors)
 
--- a/test/create-stream-vs-put-racecondition.js
+++ b/test/create-stream-vs-put-racecondition.js
@@ -3,7 +3,7 @@
  * MIT License <https://github.com/level/levelup/blob/master/LICENSE.md>
  */
 
-var levelup = require('../lib/levelup.js')
+var levelup = require('../lib/levelup.js').default
 var common = require('./common')
 var assert = require('referee').assert
 var buster = require('bustermove')
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,8 @@
+{
+  "compilerOptions": {
+    "target": "es2015",
+    "moduleResolution": "node",
+    "checkJs": true,
+    "allowJs": true
+  }
+}
--- a/lib/levelup.d.ts
+++ b/lib/levelup.d.ts
@@ -1,0 +1,96 @@
+import { EventEmitter } from 'events';
+import * as levelerrors from 'level-errors';
+import { AbstractLevelDOWN, AbstractIteratorOptions, Batch } from 'abstract-leveldown';
+
+declare global {
+  namespace Level {
+    export interface UP<K, V, O, PO, GO, DO, IO, BO, TB> {
+    }
+  }
+}
+
+export interface LevelUp<K=any, V=any, O={}, PO={}, GO={}, DO={}, IO={}, BO={}, TB=Batch<K, V>>
+  extends Level.UP<K, V, O, PO, GO, DO, IO, BO, TB>,
+  EventEmitter {
+
+  open(): Promise<void>;
+  open(callback?: (err: any) => any): void;
+  close(): Promise<void>;
+  close(callback?: (err: any) => any): void;
+
+  put(key: K, value: V, options?: PO): Promise<void>;
+  put(key: K, value: V, options: PO, callback: (err: any) => any): void;
+  put(key: K, value: V, callback: (err: any) => any): void;
+
+  get(key: K, options?: GO): Promise<any>;
+  get(key: K, options: GO, callback: (err: any, value: any) => any): void;
+  get(key: K, callback: (err: any, value: any) => any): void;
+
+  del(key: K, options?: DO): Promise<void>
+  del(key: K, options: DO, callback: (err: any) => any): void;
+  del(key: K, callback: (err: any) => any): void;
+
+  batch(array: TB[], options?: BO): Promise<void>;
+  batch(array: TB[], options: BO, callback: (err?: any) => any): void;
+  batch(array: TB[], callback: (err?: any) => any): void;
+
+  batch(): LevelUpChain<K, V>;
+
+  isOpen(): boolean;
+  isClosed(): boolean;
+
+  createReadStream(options?: IO & AbstractIteratorOptions<K>): NodeJS.ReadableStream;
+  createKeyStream(options?: IO & AbstractIteratorOptions<K>): NodeJS.ReadableStream;
+  createValueStream(options?: IO & AbstractIteratorOptions<K>): NodeJS.ReadableStream;
+
+  /**emitted when a new value is 'put' */
+  on(event: 'put', cb: (key: K, value: V) => void): this
+  /**emitted when a value is deleted*/
+  on(event: 'del', cb: (key: K) => void)
+  /**emitted when a batch operation has executed */
+  on(event: 'batch', cb: (ary: any[]) => void)
+  /**emitted when the database has opened ('open' is synonym) */
+  on(event: 'ready', cb: () => void)
+  /**emitted when the database has opened */
+  on(event: 'open', cb: () => void)
+  /** emitted when the database has closed*/
+  on(event: 'closed', cb: () => void)
+  /** emitted when the database is opening */
+  on(event: 'opening', cb: () => void)
+  /** emitted when the database is closing */
+  on(event: 'closing', cb: () => void)
+}
+
+interface LevelUpConstructor {
+  <K=any, V=any, O=any, PO={}, GO={}, DO={}, IO={}, BO={}, B = Batch<K, V>>(
+    db: AbstractLevelDOWN<any, any, O, PO, GO, DO, IO, BO>,
+    options: O,
+    cb?: (err: Error) => void): LevelUp<K, V, O, PO, GO, DO, IO, BO, B>;
+  <K=any, V=any, O=any, PO={}, GO={}, DO={}, IO={}, BO={}, B = Batch<K, V>>(
+    db: AbstractLevelDOWN<any, any, O, PO, GO, DO, IO, BO>,
+    cb?: (err: Error) => void): LevelUp<K, V, O, PO, GO, DO, IO, BO, B>;
+
+  new <K=any, V=any, O=any, PO={}, GO={}, DO={}, IO={}, BO={}, B = Batch<K, V>>(
+    db: AbstractLevelDOWN<any, any, O, PO, GO, DO, IO, BO>,
+    options: O,
+    cb?: (err?: Error) => void): LevelUp<K, V, O, PO, GO, DO, IO, BO, B>;
+  new <K=any, V=any, O=any, PO={}, GO={}, DO={}, IO={}, BO={}, B = Batch<K, V>>(
+    db: AbstractLevelDOWN<any, any, O, PO, GO, DO, IO, BO>,
+    cb?: (err?: Error) => void): LevelUp<K, V, O, PO, GO, DO, IO, BO, B>;
+
+  errors: typeof levelerrors;
+}
+
+export interface LevelUpChain<K=any, V=any> {
+  readonly length: number;
+  put(key: K, value: V): this;
+  del(key: K): this;
+  clear(): this;
+  write(callback: (err?: any) => any): this;
+  write(): Promise<this>;
+}
+
+export var errors: typeof levelerrors;
+
+declare const LevelUp: LevelUpConstructor;
+export default LevelUp
--- a/test/deferred-open-test.js
+++ b/test/deferred-open-test.js
@@ -3,9 +3,9 @@
  * MIT License <https://github.com/level/levelup/blob/master/LICENSE.md>
  */
 
-var levelup = require('../lib/levelup.js')
-var leveldown = require('leveldown')
-var encDown = require('encoding-down')
+var levelup = require('../lib/levelup.js').default
+var leveldown = require('leveldown').default
+var encDown = require('encoding-down').default
 var async = require('async')
 var common = require('./common')
 var assert = require('referee').assert
--- a/test/idempotent-test.js
+++ b/test/idempotent-test.js
@@ -3,8 +3,8 @@
  * MIT License <https://github.com/level/levelup/blob/master/LICENSE.md>
  */
 
-var levelup = require('../lib/levelup.js')
-var leveldown = require('leveldown')
+var levelup = require('../lib/levelup.js').default
+var leveldown = require('leveldown').default
 var common = require('./common')
 var assert = require('referee').assert
 var buster = require('bustermove')
--- a/test/init-test.js
+++ b/test/init-test.js
@@ -3,8 +3,8 @@
  * MIT License <https://github.com/level/levelup/blob/master/LICENSE.md>
  */
 
-var levelup = require('../lib/levelup.js')
-var leveldown = require('leveldown')
+var levelup = require('../lib/levelup.js').default
+var leveldown = require('leveldown').default
 var common = require('./common')
 var assert = require('referee').assert
 var refute = require('referee').refute
--- a/test/inject-encoding-test.js
+++ b/test/inject-encoding-test.js
@@ -3,9 +3,9 @@
  * MIT License <https://github.com/level/levelup/blob/master/LICENSE.md>
  */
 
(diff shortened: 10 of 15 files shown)


package.json
```diff
@@ -33,35 +33,39 @@
...
     "slow-stream": "0.0.4",
     "standard": "^10.0.2",
-    "tape": "^4.7.0"
+    "tape": "^4.7.0",
+    "ts-node": "^3.3.0",
+    "typescript": "^2.5.2"
   },
   "scripts": {
-    "test": "standard && node test | faucet" 
+    "test": "standard && ts-node --no-cache test | faucet"
```
vweevers: I think it should be *in addition to* node tests. Tests passing in TS is not a guarantee for tests passing in node.

test/common.js
```diff
@@ -12,13 +12,15 @@ var rimraf = require('rimraf')
...
+var leveluperrors = require('../lib/levelup.js').errors
 var errors = require('level-errors')
 var dbidx = 0
-var leveldown = require('leveldown')
-var encDown = require('encoding-down')
+var leveldown = require('leveldown').default
+var encDown = require('encoding-down').default
 
-assert(levelup.errors === errors)
+assert(leveluperrors === errors)
```
vweevers: Is this a requirement for TS? I like the previous better

MeirionHughes: unfortunately, without moving the test to ts (so you can trick the compiler) I'm not sure. the alternative would be to have specific typescript-based tests for common things - that way the typescript stuff would not interfere at all with the js tests. I don't mind doing that with something like mocha.

tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "compilerOptions": {
+    "target": "es2015",
+    "moduleResolution": "node",
+    "noEmitOnError": true,
```
vweevers: Didn't notice this before. What does `noEmitOnError` mean?

MeirionHughes: Generally a fail-safe to ensure it doesn't just warn then run, but fails completely. Its probably not needed as it defiantly was failing the test before without it.

package.json
MeirionHughes: Sure, I can switch back to having both.

test/common.js
MeirionHughes: I'll have another stab; might be a way to add the errors onto the default too.

package.json
```diff
@@ -33,35 +33,39 @@
     "storage",
     "json"
   ],
-  "main": "lib/levelup.js",  
+  "main": "lib/levelup.js",
+  "typings": "lib/levelup.js",
```
vweevers: Shouldn't this be `lib/levelup.d.ts`?

tsconfig.json
vweevers: OK so it's just a confusing name :)

test/create-stream-vs-put-racecondition.js
```diff
@@ -3,7 +3,7 @@
  * MIT License <https://github.com/level/levelup/blob/master/LICENSE.md>
  */
 
-var levelup = require('../lib/levelup.js')
+var levelup = require('../lib/levelup.js').default
```
zixia: I believe this is not necessary because we are both had exported `LevelUP` and `default`.

See: https://github.com/Level/levelup/blob/f93790fc9d296d6be0691d52687f0c204f116473/lib/levelup.js#L294

That will make sure the code will pass the following assertion:
```
const levelup1 = require('../lib/levelup.js')
const levelup2 = require('../lib/levelup.js').default

assert(levelup1 === levelup2)
```

MeirionHughes: Its down to typescript - you cannot have both default and `export = ` in the typings. So the typings have gone with the `default` export exclusively. It means when you run the normal js through the typescript compiler it will complain if you don't use the default.

zixia: Ok, understand.

Then how about we do a `module.exports = LevelUP.default = LevelUP.LevelUP = Levelup` ?

Then we will be able to do a `var { LevelUP } = require('../lib/levelup.js')`, which would be better than `require('...').default`?

vweevers: I'm not a fan of `require('...').default` - but because it's only in the tests, I prefer it over more hacks in `lib/levelup.js`.

MeirionHughes: yeah I don't think we need more hacks to the export; Its just for the test here because we're pushing the (mostly) unadulterated js through the compiler. 

you can always do const `{default:levelup} =require('levelup');` but I suspect we might run into issues if we're still testing on an older node.

zixia: I'm ok with `default` too because it's just unit test script.

README.md
```diff
@@ -24,6 +24,7 @@ LevelUP
   * <a href="#events">Events</a>
   * <a href="#extending">Extending LevelUP</a>
   * <a href="#multiproc">Multi-process access</a>
+  * <a href="#typings">Typescript </a>
```
vweevers: Nitpick: stylize like the official "TypeScript"? Same below

README.md
```diff
@@ -505,6 +506,15 @@ LevelDB is thread-safe but is **not** suitable for accessing with multiple proce
 
 See the <a href="https://github.com/level/levelup/wiki/Modules"><b>wiki</b></a> for some LevelUP extensions, including [multilevel](https://github.com/juliangruber/multilevel), that may help if you require a single store to be shared across processes.
 
+<a name="typings"></a>
+Typescript 
+----------
+
+LevelUP comes with typescript definitions that can automatically infer options based on a typed `abstract-leveldown` implementation.
```
vweevers: Maybe the simpler form: "infer options from a typed"

MeirionHughes: done

MeirionHughes:
@zixia any chance you could [pull/fork from me](https://github.com/MeirionHughes/levelup.git) and test this locally?

if you do 
```
yarn
yarn link
```
in the the local `levelup` folder it will be linkable in your own project. There you would use `yarn link levelup`. It would be super useful to make sure it all types correctly, for someone else.

MeirionHughes:
@vweevers yeah, I'll squash and rebase after #498

vweevers:
Does the [wiki page](https://github.com/Level/levelup/wiki/Typings) need updating?

zixia:
No problem, will do it later today.

vweevers:
No need, looks like @MeirionHughes already did

zixia:
@MeirionHughes I use your repository to install `levelup`, and everything looks good.

It passed my `tslint`, `rollup`(by npm run dist) and passed all my unit tests as well.

My commit to replace `levelup` to your git repo: https://github.com/zixia/flash-store/commit/c9e12531b1eddeddb6201ec5196f072f70f35e33

P.S. I think we should consider to create a new repository for TypeScript typing testing, because all the `level-*` modules sometimes might run into typing mismatch problem. It should be tested automatically.

<details>
<summary>
See the details:
</summary>

```shell
┌ zixia@zixia-desktop:~/git/flash-store [18:33:10] tty:[0] jobs:[0]
└ (MeirionHughes) $ grep MeirionHughes package.json 
    "levelup": "MeirionHughes/levelup",
┌ zixia@zixia-desktop:~/git/flash-store [18:33:13] tty:[0] jobs:[0]
└ (MeirionHughes) $ npm run lint

> flash-store@0.0.20 lint /home/zixia/git/flash-store
> tslint --project tsconfig.json && npm run clean && tsc --noEmit


> flash-store@0.0.20 clean /home/zixia/git/flash-store
> shx rm -fr dist/* bundles/*

┌ zixia@zixia-desktop:~/git/flash-store [18:33:18] tty:[0] jobs:[0]
└ (MeirionHughes) $ npm run dist

> flash-store@0.0.20 dist /home/zixia/git/flash-store
> npm run clean && npm run build && npm run rollup && npm run dist:es6to5


> flash-store@0.0.20 clean /home/zixia/git/flash-store
> shx rm -fr dist/* bundles/*


> flash-store@0.0.20 build /home/zixia/git/flash-store
> tsc --module es6


> flash-store@0.0.20 rollup /home/zixia/git/flash-store
> rollup -c

(!) Some options have been renamed
https://gist.github.com/Rich-Harris/d472c50732dab03efeb37472b08a3f32
options.entry is now options.input
options.moduleName is now options.name
options.sourceMap is now options.sourcemap
options.dest is now options.output.file
options.format is now options.output.format

dist/flash-store.js → bundles/flash-store.es6.umd.js...
(!) Mixing named and default exports
Consumers of your bundle will have to use bundle['default'] to access the default export, which may not be what you want. Use `exports: 'named'` to disable this warning
(!) Missing shims for Node.js built-ins
Creating a browser bundle that depends on 'path'. You might need to include https://www.npmjs.com/package/rollup-plugin-node-builtins
(!) `this` has been rewritten to `undefined`
https://github.com/rollup/rollup/wiki/Troubleshooting#this-is-undefined
dist/flash-store.js
1: var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
                    ^
2:     return new (P || (P = Promise))(function (resolve, reject) {
3:         function fulfilled(value) { try { step(generator.next(value)); } catch (e) { reject(e); } }
...and 9 other occurrences
(!) Unresolved dependencies
https://github.com/rollup/rollup/wiki/Troubleshooting#treating-module-as-external-dependency
path (imported by dist/flash-store.js)
app-root-path (imported by dist/flash-store.js)
brolog (imported by dist/flash-store.js)
rimraf (imported by dist/flash-store.js)
encoding-down (imported by dist/flash-store.js)
leveldown (imported by dist/flash-store.js)
levelup (imported by dist/flash-store.js)
(!) Missing global variable names
Use options.globals to specify browser global variable names corresponding to external modules
path (guessing 'path')
app-root-path (guessing 'appRootPath')
brolog (guessing 'brolog')
rimraf (guessing 'rimrafProxy')
encoding-down (guessing 'encodingProxy')
leveldown (guessing 'leveldownProxy')
levelup (guessing 'levelupProxy')
created bundles/flash-store.es6.umd.js in 45ms

> flash-store@0.0.20 dist:es6to5 /home/zixia/git/flash-store
> tsc --out ./bundles/flash-store.umd.js --target es5 --allowJs bundles/flash-store.es6.umd.js --lib es6,dom

┌ zixia@zixia-desktop:~/git/flash-store [18:33:23] tty:[0] jobs:[0]
└ (MeirionHughes) $ npm t

> flash-store@0.0.20 test /home/zixia/git/flash-store
> npm run lint && npm run test:unit


> flash-store@0.0.20 lint /home/zixia/git/flash-store
> tslint --project tsconfig.json && npm run clean && tsc --noEmit


> flash-store@0.0.20 clean /home/zixia/git/flash-store
> shx rm -fr dist/* bundles/*


> flash-store@0.0.20 test:unit /home/zixia/git/flash-store
> blue-tape -r ts-node/register -r source-map-support/register "src/**/*.spec.ts" "tests/**/*.spec.ts"

TAP version 13
# constructor()
ok 1 should not throw exception with a non existing workDir
# Store as iterator
# async iterator for empty store
ok 2 should create the workDir
ok 3 should get empty iterator
# async iterator
ok 4 should get key back
ok 5 should get val back
ok 6 should iterate once
# get()
# return null for non existing key
ok 7 should get null for not exist key
# store string key/val
ok 8 should get VAL after set KEY
# store object value
ok 9 should get VAL_OBJ after set KEY
# put()
ok 10 should put VAL for KEY
# count()
ok 11 should get count 0 after init
ok 12 should get count 1 after put
# keys()
ok 13 should get 0 key after init
ok 14 should get back the key
ok 15 should get 1 key after 1 put
# values()
ok 16 should get 0 value after init
ok 17 should get back the value
ok 18 should get 1 value after 1 put

1..18
# tests 18
# pass  18

# ok
```
</details>

MeirionHughes:
@zixia Nice. thank you. I'll wait on #498 then and do the final commit.

ralphtheninja:
@zixia How do you make those "details"? Just type opening and closing tags manually or do you use some sort of tool for it?

zixia:
@ralphtheninja Ha, I learned this trick from someone else on Github before...

I guess it's a html5 tag which will be parsed by Github:

```html
<details>
<summary>
See the details:
</summary>
bla...
bla...
bla...
</details>
```

ralphtheninja:
@MeirionHughes Mind rebasing this onto latest master?

MeirionHughes:
/sigh... like wack-o-mole these typings... 

image

typescript is now pulling up the leveldown iterator options as the actual `IO` generic. I'll see if I can fix it.

vweevers:
@MeirionHughes this does sound like we should have continuous integration tests, as @zixia suggested. Or do you think that's overkill and that the typings will become stable soon?

MeirionHughes:
A separate repo for typescript testing - to be used when anyone PRs to `level-x` sounds difficult to set up so it gets the right sources. 

I think it more feasible to have one repo for _most_ `level-x` typings, test there, then have each level-x depend on that for development. But this brings issues with package versioning. 

As for the current typing and this PR - I thought I'd managed to the get the iterator generic working perfectly, but somehow its now pulling up that extra type union (see above). While its not perfect, it still does default to `any` for the iterator options (`lte, gte, etc...`) when you use it at levelup. For the sake of my sanity, it might just be easier to drop the Key/Value generics on the iterator. 

maybe @zixia fancies another stab at typings. Get a fresh perspective.

vweevers:
> A separate repo for typescript testing - to be used when anyone PRs to level-x sounds difficult to set up so it gets the right sources.

There are tools that can help with this. It's also fine to start off with something that requires manually triggering builds. Esp. because the alternative (all typings in one repo), like you say, is more difficult to maintain with regards to versioning.

> For the sake of my sanity, it might just be easier to drop the Key/Value generics on the iterator.
> maybe @zixia fancies another stab at typings. Get a fresh perspective.

Sounds like a plan. You could also open an issue for this and tag it with `help wanted`.

MeirionHughes:
I think this is as ready as it'll ever be, without dropping the option generics entirely. The inferring of the iterator options isn't perfect, but it does default to `any`. If you explicitly type out the `Levelup< ... >` type generics you're using, its perfect.

MeirionHughes:
Need to make a decision on this because typescript users are having issues. 

There is one minor imperfection with regard to the _inferring_ of iterator options (typed). If you _explicitly_ type out the LevelUp type manually there is no issue. That said, these generics are insane... I think there are 9 generic types on LevelUp. 

An alternative would be to strip the generics entirely, remove support for infering from the given AbstractLevelDown and type LevelUP with a very limited set of _very_ common options, while allowing other (untyped) options to be passed, where applicable. 

[](https://m131jyck4m.execute-api.us-west-2.amazonaws.com/prod/poll/01BX6XSM00CG8HRNQNN29ZPPN5/Current%20-%20Can%20infer%20most%20Options%20from%20Down%20except%20iterator%20types.%20%20/vote)
[](https://m131jyck4m.execute-api.us-west-2.amazonaws.com/prod/poll/01BX6XSM00CG8HRNQNN29ZPPN5/Current%20but%20remove%20Remove%20type%20generics%20from%20iterator%20options/vote)
[](https://m131jyck4m.execute-api.us-west-2.amazonaws.com/prod/poll/01BX6XSM00CG8HRNQNN29ZPPN5/Drop%20option%20generics%20entirely%20-%20don't%20type%20options/vote)
[](https://m131jyck4m.execute-api.us-west-2.amazonaws.com/prod/poll/01BX6XSM00CG8HRNQNN29ZPPN5/Drop%20option%20generics%20entirely%20-%20keep%20common%20options%20only.%20/vote)

vweevers:
@MeirionHughes what do you mean by common options? What are those?

MeirionHughes:
Any option that is the same across all abstract-level-down implementations.

vweevers:
I don't feel qualified to vote. Defer to your judgement on this.

juliangruber:
> I don't feel qualified to vote. Defer to your judgement on this.

me neither. Which is why I'm starting to be quite sure that we shouldn't land typing without a team that maintains them, ie more than one person who is on board and able to commit.

zixia:
My two cents is to landing the current typing definition, and then we will be able to maintain it.

I'm currently using the following dependencies for my TypeScript project:

```json
    "encoding-down": "^2.3.1",
    "leveldown": "^2.0.0",
    "levelup": "github:MeirionHughes/levelup",
```

juliangruber:
What though if we need to do a major types change?

For now, we could merge it but say they are unstable and might break without a major release.

zixia:
@juliangruber I'll be ok with "the typings are unstable and might break without a major release" because:
1. it will only affect the TS developer
1. it's better than nothing
1. it will not affect the JS developer at all.

juliangruber:
> it will only affect the TS developer

It would be great to make TS developers real first class citizens for leveldb, and that's why we need to get the integration right from the start.

Is everyone ok with following a stability index as nodejs does for the typings, and mark them experimental at first?

ralphtheninja:
> Is everyone ok with following a stability index as nodejs does for the typings, and mark them experimental at first?

Sounds good to me!

MeirionHughes:
@sandersn Thanks for the fixes on abstract-leveldown; would welcome any thoughts you had on the `levelup` typings here. The general idea was for levelup to infer option types from the supplied `abstract-leveldown` implementation. But as you can see the generics are somewhat complicated.

vweevers:
What needs to happen here? Add a note to the readme about the typing stability, rebase, and merge?

@MeirionHughes

MeirionHughes:
yeah probably; its been stalled because there is no right answer at the moment; the complexity of typing the high-level `levelup` to whatever low-level `leveldown` is given to it just makes this difficult the pros and cons to any approach.  That problem simply goes away entirely in the even higher `level` project though.  Even the poll doesn't help really as its almost split equally. :D I'm probably not going to be using this for another 3/4 months as I've been distracted on another project for the last few months. I'd lean to just getting it in as is IFF @zixia is happy with it still as he/they has used it more.

vweevers:
Then let's get it in as-is. Perhaps usage in the wild will lead to more feedback.

Do you have the time or do you want someone to take over?

MeirionHughes:
@vweevers yes, if possible. Spinning too many plates at the moment to give this the attention it deserves.

zixia:
@MeirionHughes @vweevers I vote for getting it in as-it-is because as @MeirionHughes said, I'm happy with it. :-p

vweevers:
> Is everyone ok with following a stability index as nodejs does for the typings, and mark them experimental at first?

So how do we achieve this?

[](https://api.gh-polls.com/poll/01C0NK44CPQJDJXKH7SGXAMKA5/add%20%22typescript%3A%20experimental%22%20badge%20next%20to%20other%20badges/vote)
[](https://api.gh-polls.com/poll/01C0NK44CPQJDJXKH7SGXAMKA5/add%20%22stability%3A%20experimental%22%20badge%20to%20TypeScript%20section/vote)
[](https://api.gh-polls.com/poll/01C0NK44CPQJDJXKH7SGXAMKA5/add%20%22stability%3A%20experimental%22%20badge%20to%20wiki%20page/vote)

Optional additions:

[](https://api.gh-polls.com/poll/01C0NKPT5XF7MBYX30WR88MFV1/link%20the%20badge%20to%20Node.js%20stability%20index%20docs/vote)
[](https://api.gh-polls.com/poll/01C0NKPT5XF7MBYX30WR88MFV1/explain%20it%20in%20words/vote)
[)](https://api.gh-polls.com/poll/01C0NKPT5XF7MBYX30WR88MFV1/recruit%20more%20ts%20maintainers%20(with%20link%20to%20issue)/vote)

Examples:

[experimental](https://nodejs.org/api/documentation.html#documentation_stability_index) [experimental](https://nodejs.org/api/documentation.html#documentation_stability_index)

sqwk:
Is there something inherently wrong with the types that is keeping this from being merged? Been using MeirionHughes' repo for the last month and had no problems so far.

zixia:
I'd like to suggest to merge the types this month instead of using the repo from MeirionHughes.

@Level/typescript

MeirionHughes:
I'll review this as soon as I can and hopefully pull the trigger. 

There is an alternative where we create "fake" properties to carry the option types along. They wouldn't existed in the run-time: 

image

MeirionHughes:
An insane alternative to using generics... 

```ts
import {Omit} from 'type-zoo';

type unknown = {} | undefined | null;

interface AbstractOptions {
  GO: unknown;
  K: unknown;
  V: unknown;
  GET: unknown;
}

interface AbstractLevelDown extends AbstractOptions {
  GO: {
    keyAsBuffer?: boolean;
    valueAsBuffer?: boolean;
  },
  K: string | Buffer;
  V: string | Buffer;
}

interface CodecEncoder<T=unknown> {
  T: T;
  encode: (val: T) => Buffer;
  decode: (val: Buffer) => T;
  buffer: boolean
  type: string
}

interface Codec<K=unknown, V=unknown> {
  K: K;
  V: V;
  keyEncoding?: CodecEncoder<K>
  valueEncoding?: CodecEncoder<V>
}


interface EncodingDown<O extends AbstractOptions, C extends Codec> extends AbstractOptions {
  C: C;
  K: C["K"];
  V: C["V"];

  GO: O["GO"] & {
    valueAsBuffer?: boolean,
    valueEncoding?: CodecEncoder
  }

  GET(key: this["C"]["K"], cb: (value: this["C"]["V"]) => void);
  GET(key: this["C"]["K"], options: this["GO"] & { valueAsBuffer: false }, cb: (value: this["C"]["V"]) => void);
  GET(key: this["C"]["K"], options: this["GO"] & { valueAsBuffer: true }, cb: (value: Buffer) => void);
  GET<VE extends CodecEncoder>(key: this["C"]["K"], options: this["GO"] | { valueEncoding: VE }, cb: (value: VE["T"]) => void);
}

interface LevelUp<AD extends AbstractOptions> {
  GO: AD["GO"];
  K: AD["K"];
  V: AD["V"];

  GET(key: this["K"]): Promise<this["V"]>;
  GET<V = this["V"]>(key: this["K"], options: this["GO"]): Promise<this["V"]>;

  get: AD["GET"] & this["GET"];
}


declare function encoding<O extends AbstractOptions, C extends Codec>(db: O, c: C): EncodingDown<O, C>
declare function levelup<O extends AbstractOptions>(db: O): Pick<LevelUp<O>, "get">

type MyKey = {
  foo: string;
}
type MyValue = {
  bar: string;
}
type AltValue = {
  ray: string;
}

let codex = {} as Codec<MyKey, MyValue>
let encoder = encoding({} as AbstractLevelDown, codex);
let db = levelup(encoder);

async function main() {

  const alt = {} as CodecEncoder<AltValue>;

  db.get({ foo: "hello" }, (v) => {
    // v is MyValue

  });
  db.get({ foo: "hello" }, { valueEncoding: alt }, (v) => {
    // v is AltValue

  });
  db.get({ foo: "hello" }, { valueAsBuffer: true }, (v) => {
    // v is Buffer
  });

  let a = await db.get({ foo: "hello" }); // a is MyValue
  let b = await db.get({ foo: "hello" }, { valueEncoding: alt }); // b is still MyValue :(
}
```

... yeah

MeirionHughes:
Some good news at least is that type inference is dropping in typescript 2.8 ( missed this as I thought we'd only be getting conditional types). https://github.com/Microsoft/TypeScript/pull/21496

So in typescript 2.8 we should hopefully be able to just infer options and return types from the enclosed db.

Tapppi:
Figuring out what to do for level typings was a fricking pain of muddying through the different repos and issues. This should really get merged, if only to make it clear that there ARE working typings and that they are not stable.. I couldn't find working typings for the promisified version published.

vweevers:
We're considering moving the typings out of Level and into DefinitelyTyped. Please see https://github.com/Level/community/issues/16.

PR #508, opened 2017-10-08
Test: queued operation is not serialized

vweevers:
@ralphtheninja anything else we should test?

### DIFF
--- a/test/deferred-open-test.js
+++ b/test/deferred-open-test.js
@@ -7,6 +7,7 @@ var levelup = require('../lib/levelup.js')
 var leveldown = require('leveldown')
 var encDown = require('encoding-down')
 var async = require('async')
+var concat = require('concat-stream')
 var common = require('./common')
 var assert = require('referee').assert
 var refute = require('referee').refute
@@ -171,7 +172,7 @@ buster.testCase('Deferred open()', {
     }
   },
 
-  'queued operation is not serialized': function (done) {
+  'value of queued operation is not serialized': function (done) {
     var location = common.nextLocation()
     var db = levelup(encDown(leveldown(location), { valueEncoding: 'json' }))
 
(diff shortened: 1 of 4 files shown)


vweevers:
Perhaps a stream + json encoding?

test/json-test.js
```diff
@@ -7,18 +7,20 @@ var levelup = require('../lib/levelup.js')
...
 
-buster.testCase('JSON API', {
+buster.testCase('JSON encoding', {
   'setUp': function (done) {
     common.commonSetUp.call(this, function () {
       this.runTest = function (testData, assertType, done) {
         var location = common.nextLocation()
         this.cleanupDirs.push(location)
         levelup(encDown(leveldown(location), {
+          keyEncoding: 'json',
```
vweevers: Thanks to the added stream test, I found out that these tests were actually using serialized keys (`'[object Object]'`).

(this is not related to `deferred-leveldown`)

test/deferred-open-test.js
```diff
@@ -169,5 +169,24 @@ buster.testCase('Deferred open()', {
         }
       })
     }
+  },
+
+  'queued operation is not serialized': function (done) {
+    var location = common.nextLocation()
+    var db = levelup(encDown(leveldown(location), { valueEncoding: 'json' }))
```
ralphtheninja: You could do `{ keyEncoding: 'json' }` and also test the key

vweevers: Added a test.

test/deferred-open-test.js
```diff
@@ -188,5 +189,23 @@ buster.testCase('Deferred open()', {
...
+
+    this.closeableDatabases.push(db)
+    this.cleanupDirs.push(location)
+
+    // deferred-leveldown < 2.0.2 would serialize the key to a string.
+    db.put({ thing: 2 }, 'value', function (err) {
+      refute(err)
+
+      db.createKeyStream().pipe(concat(function (result) {
+        assert.equals(result, [{ thing: 2 }])
```
ralphtheninja: :+1:

PR #511, opened 2017-10-10
remove commits from changelog, moving log to releases instead

ralphtheninja:
See https://github.com/Level/levelup/releases/tag/v2.0.0


vweevers:
No commit log (anywhere) would also be fine, I think. The purpose of the changelog, whether it's in `CHANGELOG.md` or on the GH release page, is mainly to summarize high-level changes.

PR #514, opened 2017-10-16
clarify that options are specific to the underlying store

ralphtheninja:


README.md
```diff
@@ -103,7 +103,7 @@ db.put('name', 'levelup', function (err) {
 The main entry point for creating a new `levelup` instance.
 
 - `db` must be an [`abstract-leveldown`](https://github.com/level/abstract-leveldown) compliant store.
-- `options` is passed on to the underlying store.
+- `options` is passed on to the underlying store when opened and is specific to the type of store being used
```
vweevers: hmm, I just realized we have two ways of specifying options:

1. down constructor: `require('encoding-down')({ valueEncoding: 'json' })`
2. level(up) constructor -> passed to underlying storage (e.g. the binding of leveldown)

Do we need to explain this as well? (and ultimately, do we even need both?)

ralphtheninja: From an api perspective there's only one way to specify options for `level`/`levelup` but they mean different things I guess.

* In the `level` case then `options` are passed to both `encoding-down` and `leveldown`.
* In the `levelup` case then `options` are only passed on to whatever down is used.

So maybe we should make that clearer?

vweevers: Perhaps code examples are more valuable. We could add short snippets for a few distinct use cases such as `level(up)` + json encoding and `leveldown` + `{ compression: false }`.

vweevers: With the examples we now have in `encoding-down`, I don't think it's necessary to add examples here. We could however move them to `awesome` and/or the website at some point, to have a central place for integration recipes.

PR #530, opened 2017-12-26
esify lib code

aks-:
I have converted lib files to use es6 modules, left out levelup.js for now, because it would be kind of major bump. If we convert that into class, it would change how it's called hence changing api. Should I convert the test in the same PR?


ralphtheninja:
This is already a major bump since we can't support node 4.

aks-:
cool, will work on this next holiday

ljharb:
@ralphtheninja why can't you support node 4? It's still LTS until April.

ralphtheninja:
@ljharb With this change we can't support node 4 since it contains destructuring.

ljharb:
Well sure, but the prudent course of action then is either to "not use destructuring", or to add babel (which i'd recommend).

ralphtheninja:
> Well sure, but the prudent course of action then is either to "not use destructuring", or to add babel (which i'd recommend).

I agree :) I personally prefer to not use destructuring.

ralphtheninja:
I think rewriting to use `class` rather than `prototype` is unnecessary. The prototype code works well and I'm worried this will have a performance effect.

Also, my point in not supporting node 4 was that we still want to support node 4. The solution is not to remove node 4 from travis, but rather to not use destructuring.

ralphtheninja:
@vweevers This will be a problem testing on IE10, right?

vweevers:
Yes. IMO we should just keep things as-is.

PR #534, opened 2018-01-24
use markdown links to get rid of <a href> and <a name>

ralphtheninja:
I think this makes the `README` a little bit cleaner and more readable. Also easier to maintain if new sections are added.

The only thing I'm not happy with is that I can't get links working for sections with multiple words, so changed `Supported Platforms` to `Platforms`, `Promise Support` to `Promises` etc.

@vweevers Know any l337 markdown tricks to make that work?

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,13 +9,13 @@
 [![npm](https://img.shields.io/npm/dm/levelup.svg)](https://www.npmjs.com/package/levelup)
 
   * [Introduction](#introduction)
-  * [Platforms](#platforms)
+  * [Supported Platforms](#supported-platforms)
   * [Usage](#usage)
   * [API](#api)
-  * [Promises](#promises)
+  * [Promise Support](#promise-support)
   * [Events](#events)
   * [Extending](#extending)
-  * [Multi-process](#multi-process)
+  * [Multi-process Access](#multi-process-access)
   * [Support](#support)
   * [Contributing](#contributing)
   * [License](#license)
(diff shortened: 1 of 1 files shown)


doowb:
@ralphtheninja you should be able to use hyphens in place of the spaces. That's how GitHub automatically "slugifies" the headings.

vweevers:
> you should be able to use hyphens in place of the spaces. That's how GitHub automatically "slugifies" the headings.

If that also works on npm pages I prefer it over the single-word-headers.

README.md
```diff
@@ -71,19 +71,19 @@ db.put('name', 'levelup', function (err) {
...
-  * <a href="#get"><code>db.<b>get()</b></code></a>
-  * <a href="#del"><code>db.<b>del()</b></code></a>
-  * <a href="#batch"><code>db.<b>batch()</b></code> *(array form)*</a>
-  * <a href="#batch_chained"><code>db.<b>batch()</b></code> *(chained form)*</a>
-  * <a href="#isOpen"><code>db.<b>isOpen()</b></code></a>
-  * <a href="#isClosed"><code>db.<b>isClosed()</b></code></a>
-  * <a href="#createReadStream"><code>db.<b>createReadStream()</b></code></a>
-  * <a href="#createKeyStream"><code>db.<b>createKeyStream()</b></code></a>
-  * <a href="#createValueStream"><code>db.<b>createValueStream()</b></code></a>
+  * [<code><b>levelup()</b></code>](#ctor)
```
ralphtheninja: We could omit the `<a name>` below for these links as well, but I think it makes more sense to keep them as they are since they get more or less ridiculously long.

Issue #524, opened 2017-11-25
Improve "Must provide db" error message 


vweevers:
The `must provide db` error could be more helpful. Esp. if someone is expecting `levelup` to work like v1, it's not clear what "db" means, and could be misunderstood as the location of the LevelDB store.

Perhaps something like `first argument must be an abstract-leveldown compliant store`.

ralphtheninja:
:+1:

PR #539, opened 2018-02-02
improve 'must provide db' error message to make it clearer

adityapurwa:
As mentioned on #524 - I changed it to make the error message clearer.

Previously was **must provide db**,
changed to **The first `db` argument must be an abstract-leveldown compliant store**.

### DIFF
--- a/lib/levelup.js
+++ b/lib/levelup.js
@@ -46,7 +46,7 @@ function LevelUP (db, options, callback) {
   options = options || {}
 
   if (!db || typeof db !== 'object') {
-    error = new InitializationError('The first `db` argument must be an abstract-leveldown compliant store')
+    error = new InitializationError('First argument must be an abstract-leveldown compliant store')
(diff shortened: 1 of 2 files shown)


lib/levelup.js
```diff
@@ -46,7 +46,7 @@ function LevelUP (db, options, callback) {
   options = options || {}
 
   if (!db || typeof db !== 'object') {
-    error = new InitializationError('Must provide db')
+    error = new InitializationError('The first `db` argument must be an abstract-leveldown compliant store')
```
vweevers: Thanks for picking this up. I think I prefer "First argument must be.." because:

1. `db` is an internal variable, its name doesn't matter to the outside
2. "the first `db` argument" seems to imply there is a second `db` ;)

adityapurwa: You're right! Now that I reread it again it sounds weird, updating it.

Issue #550, opened 2018-02-12
rewrite test using memdown internals


ralphtheninja:
Updating to `memdown@2.0.0` will break the tests because levelup is reaching into memdown internals that have changed.

ralphtheninja:
See https://github.com/Level/levelup/issues/549#issuecomment-365070972

Issue #549, opened 2018-02-11
release levelup


ralphtheninja:
So, the big question is .. major, minor or patch? :)

ralphtheninja:
@vweevers If we look at the [Unreleased] section in the changelog, it seems to me we can release a new patch version. WDYT?

vweevers:
Due to the breaking change in `abtract-leveldown` and hence `deferred-leveldown`, doing `db.batch([null])` would theoretically throw, but it already would in `levelup` because of this code:

https://github.com/Level/levelup/blob/85d9bb104fd545738ad8d2cc7e2936c76dafb68f/lib/levelup.js#L259-L262

So yes, a patch is fine.

And when we finish #528, that can also be released as a patch, since it would still throw, just from a different place and with a different message.

vweevers:
@ralphtheninja did you look into https://github.com/Level/levelup/pull/547#issuecomment-364761938?

vweevers:
Ah that's just a test reaching into memdown internals that no longer exist (`_store`):

https://github.com/Level/levelup/blob/2f1ed67efd2430a9387c9041da4b9317f3b279ea/test/leveldown-substitution-test.js#L49-L52

So not a blocker for release.

ralphtheninja:
@vweevers We should rewrite that test though.

vweevers:
Agreed, it can use `memdown.iterator()` to collect the keys.

ralphtheninja:
@vweevers I'm a bit confused, why did that fail in the first place? The tests are passing now.

ralphtheninja:
Ignore me :smile:

ralphtheninja:
Ok, so I'll release a patch now and we can take care of #528 and the memdown.iterator() thing after that.

ralphtheninja:
2.0.2

ralphtheninja:
@vweevers I think we should remove the test using the internals of memdown. I don't see the point of it. The test is already doing `db.createReadStream()` and tests that `entries` is the same as `expected`.

ralphtheninja:
Nevermind. I wrote something anyway.

PR #551, opened 2018-02-12
remove leveldown-substitution-test.js

ralphtheninja:
~~Closes https://github.com/Level/levelup/issues/550~~


ralphtheninja:
I don't think we need to update `CHANGELOG` on each PR as long as we make sure to update it on each release. If we try to release often, it should be minimal work.

vweevers:
Re https://github.com/Level/levelup/issues/549#issuecomment-365082569:

> I think we should remove the test using the internals of memdown. I don't see the point of it. The test is already doing db.createReadStream() and tests that entries is the same as expected.

I think you're right. The point of the test *used to be* that `levelup` would default to `leveldown` as backing store. So when swapping it for `memdown`, the test had to check not only that the `levelup` instance returned the right data but that this data was actually stored in memdown.

vweevers:
Maybe we can just remove this file entirely.

ralphtheninja:
Should be fine now.

ralphtheninja:
https://travis-ci.org/Level/levelup/builds/340727414

Issue #576, opened 2018-05-23
All test/* files are published to npm


ralphtheninja:
```
test/argument-checking-test.js
test/batch-test.js
test/binary-test.js
test/browserify-test.js
test/common.js
test/create-stream-vs-put-racecondition.js
test/data/browser-throws.js
test/data/browser-works.js
test/data/testdata.bin
test/deferred-open-test.js
test/get-put-del-test.js
test/idempotent-test.js
test/index.js
test/init-test.js
test/inject-encoding-test.js
test/json-test.js
test/key-value-streams-test.js
test/maybe-error-test.js
test/no-encoding-test.js
test/null-and-undefined-test.js
test/open-patchsafe-test.js
test/read-stream-test.js
test/snapshot-test.js
test/test-10k-times.sh
```

Add to .npmignore

PR #583, opened 2018-05-29
update .npmignore

ralphtheninja:
After update:

```
lms@x260 ~/src/level/levelup (npmignore)
$ irish-pub
npm will publish levelup@3.0.1 as ralphtheninja, including the following files:

package.json
CHANGELOG.md
LICENSE.md
README.md
sauce-labs.svg
UPGRADING.md
lib/batch.js
lib/levelup.js
lib/promisify.js
```

Closes https://github.com/Level/levelup/issues/576

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,5 @@
-*.tar.gz
-build/
-test-data.tar
-test-data.db.tar
-test/benchmarks/
-level/
+test/*
+buster.js
+.airtap.yml
+.travis.yml
+.dntrc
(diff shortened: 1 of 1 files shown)


.npmignore
```diff
@@ -1,6 +1,5 @@
-*.tar.gz
-build/
-test-data.tar
-test-data.db.tar
-test/benchmarks/
-level/
\ No newline at end of file
+test/*
```
vweevers: really really nitpicking: I don't think the asterisk is necessary

Issue #603, opened 2018-07-13
Add options argument to Batch#write


vweevers:
https://github.com/Level/levelup/blob/c3845eea4547a829b05548c092b84d5cc30b8e46/lib/batch.js#L50

The argument can default to `{}` (so this is safe to release as a patch) and then passed to:

https://github.com/Level/levelup/blob/c3845eea4547a829b05548c092b84d5cc30b8e46/lib/batch.js#L61

Passing options to `write` has been supported in `abstract-leveldown` since forever: https://github.com/Level/abstract-leveldown/commit/efbd19aa59ce423e2c64be60bf29a00bb98c50ea

PR #604, opened 2018-07-13
Batch write options

ralphtheninja:
Closes https://github.com/Level/levelup/issues/603


vweevers:
Can you also add a test for options but no callback (i.e. returning a promise)?

PR #605, opened 2018-07-14
Prepare 3.1.1

ralphtheninja:


vweevers:
GH is having problems, doesn't allow posting comments on the diff. So I'll do it here:

- Are we sure the `level-iterator-stream` upgrade is safe to include in a patch?
- Can you add my name to the airtap upgrade?
- The "Call `rs.destroy()` without .." should be under Fixed

ralphtheninja:
> Are we sure the level-iterator-stream upgrade is safe to include in a patch?

Is https://github.com/Level/iterator-stream/pull/34 a breaking change?

vweevers:
I thought about https://github.com/Level/iterator-stream/pull/34#issuecomment-401098975 for a minute and I think we're good. When an error occurs, we now emit `error` and `close`. Before it was only `error`. Because the new event is emitted after the existing event, it's unlikely to break existing flows, which probably stop after an `error` anyway. More importantly, nothing happens between `error` and `close`. When we emit `error`, we've already cleaned up (ended the iterator).

ralphtheninja:
@vweevers I have pushed an update but it doesn't show in GH yet.

ralphtheninja:
> I checked it out locally, LGTM! I assume you'll update the Unreleased header before publish?

Should I be honest? :wink: Just too tired and making mistakes all the time. Will update!

PR #621, opened 2018-11-24
Replace remark with hallmark

vweevers:
Squash on merge. Closes #614, closes #615, closes #620.

The third commit adds a lot of diff noise, because `hallmark` has a custom list item indent option. Should we remove that from `hallmark`? With remark defaults, lists look like this (two spaces):

```markdown
-  foo
-  bar
```

With `hallmark` they look like (single space):

```markdown
- foo
- bar
```

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,7 +1,7 @@
 # Contributors
 
 | Name                 | GitHub                                                 | Social                                                              |
-| -------------------- | ------------------------------------------------------ | ------------------------------------------------------------------- |
+| :------------------- | :----------------------------------------------------- | :------------------------------------------------------------------ |
 | **Lars-Magnus Skog** | [**@ralphtheninja**](https://github.com/ralphtheninja) | [**@ralph@social.weho.st**](https://social.weho.st/@ralph)          |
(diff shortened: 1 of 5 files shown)


CONTRIBUTORS.md
```diff
@@ -10,7 +10,7 @@
 | **Matteo Collina**   | [**@mcollina**](https://github.com/mcollina)           | [**@matteocollina@twitter**](https://twitter.com/matteocollina)     |
 | **James Halliday**   | [**@substack**](https://github.com/substack)           | [**@substack@twitter**](https://twitter.com/substack)               |
 | **David Björklund**  | [**@kesla**](https://github.com/kesla)                 | [**@david_bjorklund@twitter**](https://twitter.com/david_bjorklund) |
-| **Jake Verbaten**    | [**@Raynos**](https://github.com/Raynos)               | [**@raynos@twitter**](https://twitter.com/raynos)                   |
+| **Raynos**           |                                                        |                                                                     |
```
vweevers: This is caused by some change between `remark-git-contributors` 0.2.1 and 0.3.0, but IMO it's not important to fix now.

PR #637, opened 2019-01-09
fixes benchmarks

morolt:


vweevers:
FWIW these benchmarks haven't been touched in a while, I'm not sure how relevant they are today.

Out of curiosity, what are you looking to benchmark, specifically? There's also [leveldown/bench](https://github.com/Level/leveldown/tree/master/bench), recently updated. We could adapt those benchmarks for other stores.

test/benchmarks/package.json
```diff
@@ -3,14 +3,17 @@
   "version": "0.0.0",
   "description": "benchmarks for LevelUP",
   "main": "index.js",
+  "scripts": {
+    "postinstall": "cd ../../ && npm install"
```
vweevers: The assumption was that the "root" is already installed (and that should be documented).

This `postinstall` step hurts flexibility (e.g. when you have linked packages), can you please remove it?

morolt: you can update document to fix it.

vweevers: Correct. So the steps are:

```
npm i
cd test/benchmarks
npm i
node .
```

PR #645, opened 2019-03-30
Prepare 4.0.1

vweevers:


vweevers:
Hold with merging; gonna include memdown (#646).

PR #651, opened 2019-05-23
Replaced explicit dependency on bash with /usr/bin/env sh

MadsAndreasenTechPeople:


vweevers:
Let's remove this old file. Can use [`test-for`](https://github.com/ralphtheninja/test-for) instead.

vweevers:
Thanks!

PR #660, opened 2019-06-25
Support options passed to .open

achingbrain:
LevelDown supports an [optional `options` argument](https://github.com/Level/abstract-leveldown#dbopenoptions-callback) passed the `db.open` method.  Currently LevelUp fails if options are passed to `.open` as it tries to invoke them as the callback so this PR adds support for that.

By and large you don't call `.open` directly but if you do (or an intermediate module does as in my case) LevelUp should probably honour the api.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ db.get('foo', function (err, value) {
 
 <a name="open"></a>
 
-### `db.open([callback])`
+### `db.open([options][, callback])`
 
 Opens the underlying store. In general you should never need to call this method directly as it's automatically called by <a href="#ctor"><code>levelup()</code></a>.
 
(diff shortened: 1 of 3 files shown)


vweevers:
Options that you pass to the `levelup` constructor are passed on to `db.open()`.

achingbrain:
> Options that you pass to the levelup constructor are passed on to db.open()

That behaviour is maintained.

> I'm hoping we can move away from db.open() options altogether

Sounds like a good plan, but if the goal here is to support the same API as `abstract-leveldown` but with added streams, then support for passing options to `db.open` should be in there until it's removed from `abstract-leveldown` itself.

vweevers:
> but if the goal here is to support the same API as abstract-leveldown but with added streams

That's not - or has not been - the goal. However there is a long-term goal to merge `levelup` functionality into `abstract-leveldown`. So on second thought, it's good to bring the two interfaces closer together :) I'm in

vweevers:
This needs a follow-up; browser tests are failing on the use of arrow functions.

vweevers:
4.1.0

PR #669, opened 2019-08-18
Add clear()

### LINKED ISSUE #648: clear() method
A DB method for removing all the keys and values. Currently doing it like:
```javascript
db.createKeyStream().on("data", key => {
    db.del(key);
});
```

vweevers:
Closes #648.

~~Blocked by https://github.com/Level/deferred-leveldown/pull/72 and https://github.com/Level/encoding-down/pull/89.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,7 +396,7 @@ Returns an [`abstract-leveldown` iterator](https://github.com/Level/abstract-lev
 
 ### `db.clear([options][, callback])`
 
-**This method is experimental. Not all stores support it yet.**
+**This method is experimental. Not all underlying stores support it yet. Consult [Level/community#79](https://github.com/Level/community/issues/79) to find out if your (combination of) dependencies support `db.clear()`.**
 
 Delete all entries or a range. Not guaranteed to be atomic. Accepts the following range options (with the same rules as on iterators):
 
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ db.put('name', 'levelup', function (err) {
 - <a href="#createKeyStream"><code>db.<b>createKeyStream()</b></code></a>
 - <a href="#createValueStream"><code>db.<b>createValueStream()</b></code></a>
 - <a href="#iterator"><code>db.<b>iterator()</b></code></a>
+- <a href="#clear"><code>db.<b>clear()</b></code></a>
 
 ### Special Notes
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -446,6 +464,7 @@ const main = async () => {
 | `put`     | Key has been updated        | `key, value` (any)   |
 | `del`     | Key has been deleted        | `key` (any)          |
 | `batch`   | Batch has executed          | `operations` (array) |
+| `clear`   | Entries were deleted        | `options` (object)   |
```
vweevers: Not 100% sure about the argument (`options`). I figured a listener might want to inspect the range options, to find out what was deleted. But that information is not exact (in contrast with the `put`, `del` and `batch` events that tell you exactly which keys were modified).

vweevers:
@ralphtheninja Before we release this I want to think about adding a way to feature-detect it.

A module like [`noffle/kappa-view-level`](https://github.com/noffle/kappa-view-level) (used in `cabal-core`) may want to use `db.clear()` for its [`clearIndex()`](https://github.com/noffle/kappa-view-level/blob/0665b631669b34b3f11935f9611a8118e5c73641/index.js#L47-L68). If `cabal-core` were to switch to `level-mem` (instead of the old `memdb` that it currently uses) then the `db` it passes into `kappa-view-level` would indeed have a `clear()` method. Problem is, both these modules support passing in a custom db.

That db may be `levelup` with `clear()`, but its underlying db may _not_ have `clear()`. So a check like `typeof db.clear === 'function'` would not suffice. Even worse, the db may be `subleveldown` with the default `clear()` of `abstract-leveldown` and delete keys outside of the sublevel range :(

ralphtheninja:
Maybe we could start by adding a `clear()` to `subleveldown` to prevent from falling back to default behavior? But I guess the damage is already done there if it already falls back to using default `clear()`.

vweevers:
Yeah, we could make a patch release on `subleveldown` to either pin `abstract-leveldown` or add a noop `_clear()`. No one is using `clear()` yet and it isn't exposed on the outer `levelup` instance yet so we have time.

vweevers:
It's time for manifests (again), I think. https://github.com/Level/community/issues/42#issuecomment-529196502

ralphtheninja:
> It's time for manifests (again), I think. [Level/community#42 (comment)](https://github.com/Level/community/issues/42#issuecomment-529196502)

I was thinking manifest just now :)

vweevers:
I propose the following timeline:

1. Patch `subleveldown`
2. Release `levelup` minor with `clear()` (floats into `level` and friends)
3. Start working on manifests, as the "next big thing". In parallel, add `clear()` to `subleveldown`

Issue #273, opened 2014-08-19
convert tests into an abstract test suite module that other levelups can use


maxogden:
It would be really useful for us over at http://github.com/maxogden/dat to be able to e.g. `require('abstract-levelup/tests')` in the same way that there is an abstract-leveldown test suite.

Publishing a test suite as a module lets multiple modules all ensure compatibility since they use the same test suite. For example, [level.js uses abstract-leveldown](https://github.com/maxogden/level.js/blob/master/test/test.js), and so does [memdown](https://github.com/rvagg/memdown/blob/master/test.js) and [leveldown](https://github.com/rvagg/node-leveldown/blob/master/test/close-test.js) and others.

I liked @rvagg's idea of a test suite as a module so much that I copied the concept but for file storage: https://github.com/maxogden/abstract-blob-store

However, the levelup tests aren't a module yet. Someone needs to do the heroic task of converting the tests to use the [tape](http://npmjs.org/tape) API and make the tests require()-able

PR #674, opened 2019-09-15
Rewrite buster tests as tape tests

### LINKED ISSUE #497: Revive create-stream-vs-put-racecondition.js test
`create-stream-vs-put-racecondition.js` isn't executed in the test because the runner looks for *-test.js.

It was the same before the changes to allow the typescript to run each test:

levelup/buster.js Line 5 in e152d83
```
tests: [ 'test/*-test.js' ],
```

If you add `-test` to the end, and use  `levelup(leveldown(location))` instead of `levelup(location)`, it fails.

### LINKED ISSUE #650: Rewrite buster tests as tape tests
image

vweevers:
Closes #650. Closes #497. WIP. Has two goals:

- Remove need for deprecated dependencies (`referee`, `samsam`)
- Align test suite with `abstract-leveldown`'s, so that they may be merged later on.


vweevers:
Next up: #273, though for a different reason: to be able to call the test suite on an `abstract-leveldown` db. That will also require further refactoring of the tests, splitting them by features as well as separating "integration tests" (that don't operate on a `testCommon.factory()` db) from the test suite.

test/create-stream-vs-put-racecondition.js
```diff
@@ -1,89 +1,69 @@
...
-    .on('data', function (data) {
-      i++
+var encdown = require('encoding-down')
+var after = require('after')
+
+module.exports = function (test, testCommon) {
+  ;[true, false].forEach(function (encode) {
+    ;[true, false].forEach(function (deferredOpen) {
+      ;[true, false].forEach(function (delayedPut) {
+        makeTest(test, encode, deferredOpen, delayedPut)
```
ralphtheninja: Nice one!

vweevers:
> Not going through this in detail

I completely understand 😉

vweevers:
Because it's a lot, I'd like to keep the commits, not squash.

PR #677, opened 2019-09-22
Create test suite

### LINKED ISSUE #273: convert tests into an abstract test suite module that other levelups can use

vweevers:
Closes #273. Doesn't fully achieve what that ticket describes (being able to call the test suite externally would require moving devDependencies to dependencies, or moving the whole thing to its own module), but there's no use case for that anymore. So, this PR is more about being able to reuse the test suite _internally_. This allows us to test feature parity between `levelup` and `abstract-leveldown` (https://github.com/Level/community/issues/58).

### DIFF
--- a/test/deferred-open-test.js
+++ b/test/deferred-open-test.js
@@ -1,10 +1,12 @@
 var each = require('async-each')
 var parallel = require('run-parallel')
 var concat = require('concat-stream')
-var sinon = require('sinon')
 var readStreamContext = require('./util/rs-context')
+var rsFactory = require('./util/rs-factory')
 
 module.exports = function (test, testCommon) {
+  var createReadStream = rsFactory(testCommon)
+
   test('deferred open(): put() and get() on new database', function (t) {
     // 1) open database without callback, opens in next tick
     var db = testCommon.factory()
@@ -19,13 +21,13 @@ module.exports = function (test, testCommon) {
       // 3) when the callbacks have returned, the database should be open and those values should be in
       //    verify that the values are there
       each([1, 2, 3], function (k, next) {
-        db.get('k' + k, function (err, v) {
+        db.get('k' + k, { asBuffer: false }, function (err, v) {
           t.ifError(err)
           t.is(v, 'v' + k)
           next()
         })
       }, function () {
-        db.get('k4', function (err) {
+        db.get('k4', { asBuffer: false }, function (err) {
           t.ok(err)
           db.close(t.end.bind(t))
         })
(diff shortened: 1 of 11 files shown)


test/deferred-open-test.js
```diff
@@ -130,12 +132,12 @@ module.exports = function (test, testCommon) {
...
   })
 
-  test('deferred open(): maxListeners warning', function (t) {
+  test('deferred open(): no maxListeners warning', function (t) {
     // 1) open database without callback, opens in next tick
     var db = testCommon.factory()
-    var stderrMock = sinon.mock(console)
+    var fail = t.fail.bind(t)
 
-    stderrMock.expects('error').never()
```
vweevers: Besides this not actually working anymore (because node emits a `warning` event rather than `error`), it caused problems when calling the test multiple times: sinon complained `console` was already wrapped.

test/self.js
```diff
@@ -0,0 +1,57 @@
...
+  deferredOpen: true,
+  promises: true,
+  streams: true,
+  encodings: false
+})
+
+// TODO to make this pass:
+// - Have abstract-leveldown use level-errors
+// - Perform type checks in same order (e.g. check key before callback)
+// - Add db.isClosed(), isOpen() to abstract-leveldown
```
vweevers: Doable, I think 😃

test/util/rs-factory.js
```diff
@@ -0,0 +1,16 @@
...
+
+var xtend = require('xtend')
+
+module.exports = function (testCommon) {
+  return function createReadStream (db, options) {
+    if (!testCommon.encodings) {
+      options = xtend(options, {
+        keyAsBuffer: false,
+        valueAsBuffer: false
+      })
```
vweevers: The tests implicitly relied on the default utf8 encoding, for the db to return strings. When the db isn't wrapped with `encoding-down`, `*asBuffer: false` achieves the same.

vweevers:
Travis error:

```
The job exceeded the maximum log length, and has been terminated.
```

That's new..

PR #678, opened 2019-09-22
Add manifest

vweevers:


lib/levelup.js
```diff
@@ -55,6 +56,23 @@ function LevelUP (db, options, callback) {
...
+    deferredOpen: true,
+    openCallback: true,
+    promises: true,
+    streams: true
+  })
+
+  Object.keys(this.supports.additionalMethods).forEach(function (method) {
+    this[method] = this[method] || function () {
+      // TODO: uhm. if the manifest is based on deferred-leveldown's,
+      // should we also call the methods via deferred-leveldown?
```
vweevers: This is tricky.

vweevers: Can work if we agree that `deferred-leveldown` itself must not add methods to its `additionalMethods`; it should inherit that from its input db (same as `levelup`).

lib/levelup.js
```diff
@@ -53,8 +54,23 @@ function LevelUP (db, options, callback) {
 
   this.options = getOptions(options)
   this._db = db
-  this.db = new DeferredLevelDOWN(db)
+  this.db = db.supports && db.supports.deferredOpen ? db : new DeferredLevelDOWN(db)
```
vweevers: Should also add this logic to `levelup#close()`, but this is just a proof of concept, I'll revert dc5cdb779592dbeca81b66d2a5305697f4ee05a1 after review (I hope you can see where I'm going with this 😃)

ralphtheninja: Hmm, don't you mean the opposite of this? I.e. `db.supports && db.supports.deferredOpen ? new DeferredLevelDOWN(db) : db`

ralphtheninja: Oh nvm. I get it now :)

ralphtheninja: E.g. if the db itself supports deferred open, there's no need in wrapping it with `DeferredLevelDOWN()`.

vweevers: ✔️

lib/levelup.js
```diff
@@ -53,8 +54,23 @@ function LevelUP (db, options, callback) {
...
+    openCallback: true,
+    promises: true,
+    streams: true
+  })
+
+  Object.keys(this.supports.additionalMethods).forEach(function (method) {
+    this[method] = this[method] || function () {
+      return this.db[method].apply(this.db, arguments)
+    }
+  }, this)
```
vweevers: @ralphtheninja Shall I keep this (L69-L73) in, or do it later? Until we have manifests in every module, this `levelup` code won't actually do anything (and can't be tested).

ralphtheninja: No opinion :)

Issue #656, opened 2019-06-09
levelup constructor without callback causes unhandled rejection


bjacokes:
Using v4.0.1 of `levelup` and `leveldown`. The following command causes an unhandled rejection:

```
levelup(leveldown('/path/to/nonexistent/file'))

UnhandledPromiseRejectionWarning: OpenError: IO error: /path/to/nonexistent/file: No such file or directory
....
    at (node:69500) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
```

This happens because the final line in the `LevelUP` constructor, `this.open(callback)`, is a floating promise.

vweevers:
Aye. The unhandled rejection is somewhat consistent with the previous behavior of emitting an error event, but the difference is you can't catch it now. Perhaps we should change this line:

https://github.com/Level/levelup/blob/c56f863e41f1b00216837963be34d4ba74458b19/lib/levelup.js#L57

To:

```js
this.open(callback || function (err {
  if (err) self.emit('error', err)
})
```

vweevers:
A PR is welcome!

bjacokes:
Your proposed change would work, but it seems (maybe?) inconsistent with this line in the docs: "Breaking change: previously, if you did not pass a callback to an async function and there was an error, `levelup` would emit an `error` event instead. This is no longer true." I also don't see any other places in the code where the `error` event is emitted.

I'm happy to open a PR for the code portion, as long as `error` is still an event that you're OK emitting, and as long as you feel that no doc changes are needed.

(FWIW, I'm not blocked by this in my own project since I switched to using the `callback` form of the constructor.)

vweevers:
> "Breaking change: previously, if you did not pass a callback to an async function and there was an error, levelup would emit an error event instead. This is no longer true."

That line applies to all methods in general. The open-callback was/is the only special case.

We can document the error event [here](https://github.com/Level/levelup#levelupdb-options-callback). E.g. "If you don't provide a callback, any read & write operations are simply queued internally until the store is fully opened, _unless it fails to open, in which case an `error` event will be emitted._"

> as long as error is still an event that you're OK emitting,

It's OK. The error event is a common pattern in Node.js as well as `levelup` historically. And it's preferable to an unhandled promise rejection, which is left hanging without crashing the process (though that behavior is gonna change in Node.js I believe).

[...]

PR #679, opened 2019-09-26
Fix floating promise in LevelUP constructor

bjacokes:
As suggested in https://github.com/Level/levelup/issues/656

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ The main entry point for creating a new `levelup` instance.
 - `db` must be an [`abstract-leveldown`](https://github.com/Level/abstract-leveldown) compliant store.
 - `options` is passed on to the underlying store when opened and is specific to the type of store being used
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -470,6 +470,7 @@ const main = async () => {
 | `ready`   | Alias of `open`             | -                    |
 | `closing` | Store is closing            | -                    |
 | `closed`  | Store has closed.           | -                    |
+| `error`   | Constructor had an error    | `error` (any)        |
```
vweevers:
```suggestion
| `error`   | An error occurred           | `error`              |
```

The `(any)` text in the other cells means "any type", that's not the case here.

lib/levelup.js
```diff
@@ -54,7 +55,11 @@ function LevelUP (db, options, callback) {
   this.options = getOptions(options)
   this._db = db
   this.db = new DeferredLevelDOWN(db)
-  this.open(callback)
+  this.open(callback || function(err) {
```
vweevers:
```suggestion
  this.open(callback || function (err) {
```

(You can also run` standard --fix` to fix style issues automatically)

vweevers:
Superseded by #680. Thanks!

PR #681, opened 2019-10-04
Test manifest integration with deferred-leveldown

vweevers:


vweevers:
Note that the `deferred-leveldown` upgrade itself is semver-minor: https://github.com/Level/levelup/pull/682#issuecomment-538387804.

Issue #699, opened 2020-09-26
Refactor to use fromCallback from catering module


ralphtheninja:
That code is also tested in `catering`.

PR #700, opened 2020-09-26
Refactor promisify() code by using catering module

ralphtheninja:
Closes https://github.com/Level/levelup/issues/699


lib/levelup.js
```diff
@@ -169,17 +161,13 @@ LevelUP.prototype.isClosed = function () {
...
-  var promise
-
   callback = getCallback(options, callback)
+  callback = catering.fromCallback(callback)
 
-  if (!callback) {
-    callback = promisify()
-    promise = callback.promise
+  if (maybeError(this, callback)) {
+    return callback.promise
```
ralphtheninja, line 168: Added a few extra line breaks here since the rest of the code seems to use that style.

vweevers:
Could be considered a breaking change because we did `if (!callback)` here, while `catering` does `if (callback === undefined)`. The latter matches documented behavior though so I think it's OK. The docs never suggested we support e.g. `await db.get('foo', null)` the same as `await db.get('foo')`.

vweevers (approved):

ralphtheninja:
> Could be considered a breaking change because we did `if (!callback)` here, while `catering` does `if (callback === undefined)`. The latter matches documented behavior though so I think it's OK. The docs never suggested we support e.g. `await db.get('foo', null)` the same as `await db.get('foo')`.

Good catch. I didn't think that far. My OCD just told me to replace the code with your tested module :)

PR #716, opened 2021-04-11
Drop legacy features & runtime environments

vweevers:
Ref Level/community#98.

Blocked by release of https://github.com/Level/deferred-leveldown/pull/86, otherwise ready for review (assuming tests pass on Sauce Labs).


vweevers:
Chrome on Android 10 failed, but Android 6-9 and 11 did not, so I think we can skip Android 10 for now and look into it later.

vweevers:
[5.0.0](https://github.com/Level/levelup/releases/tag/v5.0.0)

Issue #633, opened 2018-12-31
Support encoding options on chained batch


vweevers:
While we support passing `options` to `chainedBatch.write()`, `encoding-down` encodes keys and values before `write()` is called. We can keep that as-is, but should allow encoding options on `chainedBatch.put()` and `chainedBatch.del()`.

E.g.: `chainedBatch.put(key, value, options)`. 

To be added to:

- [x] `abstract-leveldown`: [7.0.0](https://github.com/Level/abstract-leveldown/releases/tag/v7.0.0)
- [x] `encoding-down`: [7.0.0](https://github.com/Level/encoding-down/releases/tag/v7.0.0)
- [x] `levelup`: https://github.com/Level/levelup/pull/717, [5.0.0](https://github.com/Level/levelup/releases/tag/v5.0.0)
- [x] `level` (just documentation): [7.0.0](https://github.com/Level/level/releases/tag/v7.0.0)

liximomo:
Really need this

vweevers:
Done

PR #717, opened 2021-04-17
Support encoding options on chained batch put() and del()

vweevers:
Ref https://github.com/Level/levelup/issues/633


vweevers:
[5.0.0](https://github.com/Level/levelup/releases/tag/v5.0.0)

PR #718, opened 2021-04-19
chore(ci): converted travis workflow to github action

marcuslyons:
Ref [Level/community#99](https://github.com/Level/community/issues/99)


.github/workflows/test.yml
```diff
@@ -0,0 +1,32 @@
...
+          node-version: ${{ matrix.node }}
+      - name: Install
+        run: npm install
+      - name: Test
+        if: ${{ matrix.node != 14}}
+        run: npm test
+      - name: Test Browsers
+        if: ${{ matrix.node == 14}}
+        env:
+          SAUCE_CONNECT_DOWNLOAD_ON_INSTALL: true
```
vweevers, line 25: I prefer a separate workflow for the browser tests (making it easier to retry just those tests). Also needs secrets (I will create them) and adding a hostname to `hosts`.

To achieve that we can copy [`nodejs/readable-stream/.github/workflows/sauce.yml`](https://github.com/nodejs/readable-stream/blob/master/.github/workflows/sauce.yml) as-is 😃

README.md
```diff
@@ -5,7 +5,7 @@
 [![Node version](https://img.shields.io/node/v/levelup.svg)](https://www.npmjs.com/package/levelup)
 [![Travis](https://img.shields.io/travis/com/Level/levelup.svg?logo=travis&label=)](https://travis-ci.com/Level/levelup)
```
vweevers, line 6:
```suggestion
[![Test](https://github.com/Level/levelup/actions/workflows/test.yml/badge.svg)](https://github.com/Level/levelup/actions/workflows/test.yml)
```

I initially forgot this on `concat-iterator` and `supports` which I reckon you used as reference.

vweevers (changes requested):
Thank you so much and welcome! Needs two small changes.

.github/workflows/test.yml
marcuslyons: I've brought that workflow over into the PR. I think it's better for them to be separate as well.

.github/workflows/sauce.yml
```diff
@@ -0,0 +1,24 @@
+name: Sauce Labs
+on: push
```
vweevers, line 2:
```suggestion
on: push
concurrency: sauce-labs
```

See https://github.com/Level/level-js/pull/212

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Use node ${{ matrix.node }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node }}
+      - name: Install
+        run: npm install
+      - name: Test
+        if: ${{ matrix.node != 14}}
```
vweevers, line 20: This line can be removed now, right?

ralphtheninja: I don't recall what it was used for.

marcuslyons: I forgot to remove it from the previous commit.

I've taken it out now.

vweevers (approved):
I'll add the necessary secrets tomorrow and push to a branch (that has access to the secrets) so that we can see it in action. Nice work!

vweevers:
Follow-ups: https://github.com/Level/levelup/commit/521b34c77204f3d5fb390a3e3c3e617411ac541e and https://github.com/Level/levelup/commit/f4d6a1df19f8aade086feb7a226db28839656b1c

PR #721, opened 2021-05-15
chore: remove use of assert module

achingbrain:
It's a node core module that isn't declared as a dep of this module so modules depending on this one have to declare it separately when using bundlers that don't polyfill node core.

Usage of it here seems unnecessary and can be replaced with a simple equality check.


vweevers (approved):
Small note: this changes the error type from AssertionError to Error but it's highly unlikely anyone depends on that.

vweevers:
[5.0.1](https://github.com/Level/levelup/releases/tag/v5.0.1). Thanks!

PR #746, opened 2022-06-10
refactor: use `github.com/isaacs/inherits` instead of `util`

faustbrian:
Running into `Uncaught TypeError: inherits is not a function` in a project without a polyfill for `util`. https://github.com/Level/memdown uses the same dependency.

### DIFF
--- a/lib/levelup.js
+++ b/lib/levelup.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const EventEmitter = require('events').EventEmitter
-const inherits = require('util').inherits
+const inherits = require('inherits')
 const DeferredLevelDOWN = require('deferred-leveldown')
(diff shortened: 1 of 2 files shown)


vweevers:
Judging by the referenced PR, you are upgrading to latest `level`, so you should no longer need `levelup`. You can replace `memdown` with `memory-level`.

vweevers:
Unless there's an urgent need (like a vulnerability, or some fundamental inability to upgrade to `abstract-level` based modules) I prefer not to spend further time on `levelup`. It will be deprecated at some point.

vweevers:
Are you asking for before or after upgrading to `abstract-level` modules? Because if after, there should be no usage of `levelup` or `level-errors` left, and all errors should have a `code` (if not, that's a bug).

